PR #45, opened 2014-07-23
Fixed scroll action, so it doesn't scroll beyond content size

idcuesta:
**WIP** Do not merge; based of master branch.

While testing collection views, I have found a weird issue. Sometimes when scroll action scroll the collection view beyond the content size (offset + frame > contentSize) then collection view cells stop being accessible through calabash.

To fix this I have changed how scroll is performed, so if you are close to an edge of the scroll view and execute the scroll action, the calculated movement will only be until the end of the content size (instead of always scrolling half of the frame in non paged scroll views)


krukow:
Let's target this for 0.12.1

jmoody:
I really like this PR.

idcuesta:
Hi Joshua and Karl,

Sorry for having this parked for so long, I completely forgot about it :(

I have added unit tests to cover the changes made. I will work in a example of behaviour before & after the change later.

jmoody:
+1000 @idcuesta

TeresaP:
@adzellman @andyhuynh86 FYI this could enable us to use the scroll gesture on the scroll views again instead of using flick.  (We went away from it because it would scroll off screen and we had to add hacks to bring it back in view again.)

jmoody:
@TeresaP @idcuesta 

The only reason we are hesitating to merge is that we don't want to introduce breaking changes.

FYI/BTW Travis CI failed because xcodebuild could not start a simulator.  -1 Apple.

@idcuesta Have you been able to run:

```
$ scripts/test/xctest.rb
```

locally?

To run all the tests:

```
$ scripts/test/run
```

idcuesta:
* I have checked travis and it looks like a bug, it doesn't work locally for me with the script `scripts/test/xctest.rb` when it runs 
```shell
xcrun xcodebuild clean test -SYMROOT=build -derivedDataPath build -project calabash.xcodeproj -scheme calabash-ios-server-tests -destination 'platform=iOS Simulator,name=iPhone 5s,OS=latest' -sdk iphonesimulator -configuration Debug > /dev/null
```
but it works fine when I use xctool to run my tests 
```shell
xctool clean test -SYMROOT=build -project calabash.xcodeproj -scheme calabash-ios-server-tests -sdk iphonesimulator -configuration Debug
```
* When I run `scripts/test/run` it takes 20 minutes but everything seem to be fine
captura de pantalla 2015-02-17 a las 19 51 52


* If it is ok, after adding the sample behaviour for scrolling to the chou app, I'm going to create a PR into `develop`, instead of `master`. I can open it, link it to this one and close this.

jmoody:
re: xctool vs xcodebuild

Ha!  I could not get xctool to work today after a bunch of trying.  It would always launch the wrong simulator.  What was the error output of `scripts/test/xctest.rb`?

> If it is ok, after adding the sample behaviour for scrolling to the chou app, I'm going to create a PR into develop, instead of master. I can open it, link it to this one and close this.

Awesome sauce.  I want to send you ice cream.

jmoody:
It is a bug on travis that the `xcodebuild test` does not work.  I am iterating with them.

@idcuesta Be sure to pull/rebase against develop.  I am very sorry that you might have to deal with merge conflicts.

idcuesta:
Hi again,

- I have created a PR in develop, so we should be able to close this one.
- I have a PR for your smoke tests project, here -> https://github.com/jmoody/animated-happiness/pull/16, that cover the changes made in the scroll views.

I think this should be enough to demonstrate the behavior before/after the changes :)

PR #105, opened 2015-01-05
Provide a dump mechanism for webview doms

krukow:
Adds webview support to the dump route. Example output:

https://gist.github.com/krukow/6dec0732db486b548184

### DIFF
--- a/calabash/Classes/Utils/LPTouchUtils.m
+++ b/calabash/Classes/Utils/LPTouchUtils.m
@@ -301,7 +301,7 @@
 #if __IPHONE_OS_VERSION_MAX_ALLOWED >= 80000
   if ([frontWindow respondsToSelector:@selector(convertPoint:toCoordinateSpace:)]) {
     CGFloat sampleFactor = [[LPDevice sharedDevice] sampleFactor];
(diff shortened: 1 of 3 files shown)


calabash/Classes/Utils/LPTouchUtils.m
```diff
@@ -301,7 +301,7 @@ + (CGPoint) centerOfView:(UIView *) view shouldTranslate:(BOOL) shouldTranslate
 #if __IPHONE_OS_VERSION_MAX_ALLOWED >= 80000
   if ([frontWindow respondsToSelector:@selector(convertPoint:toCoordinateSpace:)]) {
     CGFloat sampleFactor = [[LPDevice sharedDevice] sampleFactor];
-    rect = [frontWindow convertRect:rect toCoordinateSpace:frontWindow];
+    rect = [window convertRect:rect toCoordinateSpace:frontWindow];
     CGFloat x = (rect.origin.x + 0.5 * rect.size.width) * sampleFactor;
```
jmoody: @krukow Was this a bug?

krukow: As far as I can see, yes. In the rate case a view is in a different window.

jmoody:
+1 for using the *feature* prefix.

PR #110, opened 2015-02-04
Release/0.12.3

jmoody:
* **Fix jsonify view crash on uislider** #109
* Added a changelog for the server

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,4 @@
+### 0.12.3
+
+* Fix jsonify view crash on uislider #109
+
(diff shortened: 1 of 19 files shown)


jmoody:
This a botched release.  Downstream integration tests revealed memory management problems and potential crashes.

This merge was tested agains the Briar iOS Example on iOS 7 and iOS 8 physical devices and passed on the iOS Simulator on Jenkins.

It was not until I ran the calabash-ios rspec tests over the course of the afternoon did I discover the issues.

The plan is to fix up XCTest environment so that it can be used to expose future memory management problems and catch changes that introduce crashes.

PR #111, opened 2015-02-05
Add app target to xctest bundle

jmoody:
#### Motivation

Integration testing of release/0.12.3 revealed a number of memory management problems and some crashes.

Adding an application target to the XCTest bundle will allow testing on physical devices and the ability to unit test the server behavior on an actual app.

The LPJSONUtils XCTests capture the current behavior of the `jsonifyView:` method.  This will be used to verify the correctness of the changes that will replace  643b976.

#### Notes
The LPTestTarget application does not link the calabash.framework.  The server sources are included in the XCTest bundle and are injected into the LPTestTarget at runtime - this is part of the XCTest (*cough*) magic.


jmoody:
WIP to resolve issues with https://github.com/calabash/calabash-ios-server/pull/110

PR #115, opened 2015-02-05
WIP fixes several examples of success checks against NSError

jmoody:
**WIP**  Expect more changes.

> Important Success or failure is indicated by the return value of the method. Although Cocoa methods that indirectly return error objects in the Cocoa error domain are guaranteed to return such objects if the method indicates failure by directly returning nil or NO, you should always check that the return value is nil or NO before attempting to do anything with the NSError object.

In this case, I don't think there is any harm, because we control the class that is being called.  However, it is not a good pattern.

Not for 0.12.3 release.


jmoody:
Will not merge. 

Replaced by: LPJSONSerializer can encode arbitrary objects #165

PR #118, opened 2015-02-17
Prefer to run cli tests with xcpretty

jmoody:
### Motivation

We've been piping to xcodebuild output to `> /dev/null` because travis does not allow the HUGE amount of output xcodebuild generates.

We would like to use xcpretty, but were having problems handling the exit code.  In particular, if there was a compile error, xcpretty would report success.

I believe this change does the right thing and forwards the correct xcodebuild exit code in all cases.


jmoody:
I forgot to mention:  I had a failing travis job today because xcodebuild could not start a simulator.  -1 Apple.

PR #120, opened 2015-02-18
Modified scrolling behaviour

idcuesta:
Changes include:
- Modified scrolling behaviour
- Added unit tests to cover the changes

### DIFF
--- a/calabash-ios-server-tests/LPScrollOperationTest.m
+++ b/calabash-ios-server-tests/LPScrollOperationTest.m
@@ -1,0 +1,328 @@
+#if ! __has_feature(objc_arc)
+#warning This file must be compiled with ARC. Use -fobjc-arc flag (or convert project to ARC).
+#endif
+
+#import <XCTest/XCTest.h>
+#import "LPScrollOperation.h"
+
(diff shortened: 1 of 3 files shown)


idcuesta:
This PR is a cleanup from the one over master branch -> https://github.com/calabash/calabash-ios-server/pull/45

Apologies for all the mess :(

jmoody:
Hurray! @idcuesta FTW!

jmoody:
@idcuesta Any chance you looked at an example that uses "pull down to refresh?"

I wonder if we have broken something here.  We don't have any examples.

idcuesta:
@jmoody I didn't try pull to refresh, by in my previous experience scrolling up was not triggering pull to refresh action anyway.

I'm now working in a sample app with a table view with refresh control, I could give it a try

jmoody:
@idcuesta 

> I'm now working in a sample app with a table view with refresh control, I could give it a try

That would be great.

FYI/BTW I am not going to delay the 0.13.0 for this.  The part of my brain that worries about breaking backward compatibility kicked in when I was writing the release notes.

PR #123, opened 2015-02-23
Feature/migrate objc code from calabash js

jmoody:
**FORCE PUSHED** Tue Feb 24 06:03 CET

@krukow 

### Motivation

* Migrates the LPWebQuery and insert-to-webview-script out of the calabash-js submodule and into this repo.

This file is now under ARC.  As such, I updated the naming conventions of methods to match the expectations for mixed ARC/non-ARC projects.  I also applied the project formatting to the file.  These are strict refactors - no behaviors were changed.

Companion to  **Prepare for migration for LPWebQuery and insert script to calabash-ios-server** [#calabash-js/7](https://github.com/calabash/calabash-js/pull/7).

The new make rule is:

```
$ make webquery_headers
```

The make rule has been added to `$ scripts/test/test-make-rules.rb` and will be run on Travis CI.

### Notes

Travis CI + XCTest is still flakey.  Seeing some of the same results locally.  Apple.  Please.

PR #126, opened 2015-02-25
Release/0.13.0

jmoody:
### 0.13.0

* Migrate objc code from calabash-js submodule #123
* Scroll gestures don't scroll past content size #120 @idcuesta
* Fix an edge-case crash in dump route #119


jmoody:
Grrr. XCTest cannot launch a simulator on Travis CI and Travis itself is having trouble updating GitHub webhooks.

PR #135, opened 2015-03-06
LPVersionRoute converted to ARC, safer fetching of values from Info.plist, and CalabashServer logs the App's Base SDK

jmoody:
This is a combination three of the behaviors.   I decided to keep them together because they all required changes to how we interact with the Info.plist.

* Put LPVersionRoute under ARC - part of an on-going effort to put the entire server under ARC.
* CalabashServer now reports the base SDK of the app under test.
* The version route also reports the base SDK of the app under test.
* The version route now reports "CFBundleVersion" vs. "CFBundleShortVersionString", which is sometimes empty.

```
 App Base SDK: iphonesimulator8.3
```

### DIFF
--- a/XCTest/LPInfoPlistTest.m
+++ b/XCTest/LPInfoPlistTest.m
@@ -58,6 +58,11 @@ describe(@"LPInfoPlist", ^{
     expect([infoPlist stringForDTSDKName]).to.equal(@"foo");
   });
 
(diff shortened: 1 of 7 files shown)


jmoody:
Looks like we need to disable some warnings until Xcode 6.3 goes live - new clang compiler version.

Also there are problems with the ARC settings on the dylibs.

PR #139, opened 2015-03-10
Maintenance patch for translating rects for legacy apps running in iOS 8 environments

jmoody:
#### Motivation

Checks for _legacy apps running in iOS 8 environments_ and assumes the responsibility of the translating the touch coordinates. 

Historically, the `Calabash::Cucumber::InstrumentsActions::normalize_rect_for_orientation!` method did this work.  Starting in 0.11.0, the server performs the normalization when the test device is iOS 8.

This patch will allow legacy apps to use the most recent version of the calabash gem (0.13.0).

A _legacy app_ has a base SDK of iOS 6.*.

@anujb @jescriba @rasmuskl @TobiasRoikjer @krukow @john7doe @acroos


jmoody:
Expect CI failures.  Travis is on the fritz again - no iOS 7.1 simulators are installed.

PR #151, opened 2015-04-26
Fix fall-through causing double write of http chunk

krukow:
There was a path through the condition route which would write an http chunk twice leading to an invalid HTTP message (causing the client parser to throw an exception).

This should be fixed with this PR.


jmoody:
@krukow There was a user who was reporting this, but I cannot find correspondence and I don't recall what channel it came in on (Forum, Stack Overflow, or GitHub issue).

jmoody:
@krukow Is there a cucumber test I can write for this?

krukow:
@jmoody I'm afraid it was a race - I'm not sure how to reproduce

PR #152, opened 2015-04-28
Guard calls to lpIsWebView since we also touch non UIView objects

krukow:
We had several call sites which called `lpIsWebView` unguarded. Since the view hierarchy can contain both DOM elements (dictionaries) as well as accessibility objects this could lead to crashes.

### DIFF
--- a/calabash/Classes/FranklyServer/Operations/LPScrollOperation.m
+++ b/calabash/Classes/FranklyServer/Operations/LPScrollOperation.m
@@ -42,7 +42,7 @@
     }
     
     return _view;
-  } else if ([_view lpIsWebView]) {
+  } else if ([_view respondsToSelector:@selector(lpIsWebView)] && [_view lpIsWebView]) {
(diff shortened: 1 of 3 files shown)


krukow:
@jmoody Note: I would rewrite the use of categories with static methods / functions to avoid this kind of problem. Also I generally prefer to avoid adding categories to any class when we can avoid it.

jmoody:
@krukow 

I think something else is wrong.  If the selector is supposed to receive a UIView and it is not receiving a UIView, then the problem is elsewhere.

krukow:
@jmoody I agree that something else is wrong. I still prefer use of functions over categories because it can be parameterized by (id) and always return in a safe way, also if the user should happen to reflect over methods of the class, ours wouldn't appear.

jmoody:
@krukow 

re:  Categories vs. functions - Gotcha and agreed.

In this case, I thought I had covered all the edge cases.  I even tested the dump route.

PR #157, opened 2015-04-30
LPSliderOperation should support any UIControlEvent the target slider is registered to receive

jmoody:
### Motivation

We got a feature request to support UIControlEvents other than UIControlEventValueDidChange for UISlider.

### TODO

- [x] @krukow This is a little beyond the scope of the original API.  Are you comfortable?

### Testing

With briar-ios-example.

### DIFF
--- a/calabash/Classes/FranklyServer/Operations/LPSliderOperation.m
+++ b/calabash/Classes/FranklyServer/Operations/LPSliderOperation.m
@@ -61,20 +61,10 @@
   [slider setValue:targetValue animated:animate];
 
   if (notifyTargets) {
-    NSSet *targets = [slider allTargets];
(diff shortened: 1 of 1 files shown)


jmoody:
@krukow Not that I am jumping at the chance to interact with UIAutomation, but the UIASlider app provides `dragToValue` which triggers UIEvents.

```
> uia("UIATarget.localTarget().frontMostApp().mainWindow().elements()[0].sliders()[0].dragToValue(0.5)")
```

This would be a +1 for replacing the LPSliderOperation with a UIAutomation API.

krukow:
@jmoody go for it. I think we should consider the UIA slider operations too though in the longer term

PR #161, opened 2015-05-02
More guards in [LPJSONUtils jsonifyAccessibilityElement:] against unrecognized selectors

jmoody:
### Motivation

Tightens up the `jsonifyAccessibilityElement:` method.

- [x] @krukow Needs a review.  It solves one class of problems, but I am still struggling with a element that has an accessibilityIdentifier that is an NSNumber.


krukow:
LGTM - merge at your leisure

PR #162, opened 2015-05-02
WIP Coerce accessibility attrs to NSString

jmoody:
**WIP DO NOT MERGE**

### Motivation

We've seen examples where users are setting Accessibility Attributes to non-NSString values.  This PR catchs exceptions that are raised when Apple's private APIs makes internal calls as the result of calling `accessibilityLabel` or `accessibilityIdentifier` on objects that have non-NSString accessibility attributes.  It also detects non-NSString ids and labels and coerces them to NSString.

#### Example

A UILabel has an accessibilityIdentifier of @(1).   This will cause calls to _accessibilityLabel_ to raise an exception.

### Discussion

I don't like the behavior that this PR introduces.  This is a bug that user should address.

-1 for merging.

### TODO

- [x] Guards added to UIScriptASTWith
- [ ] unit tests for UIScriptASTWith
- [ ] Search for other examples (there are actually not too many others).

ATTN @krukow

I made this PR because I am about to make a simulator dylib from it to distribute internally.

### DIFF
--- a/calabash/Classes/UISpec/UIScriptASTWith.m
+++ b/calabash/Classes/UISpec/UIScriptASTWith.m
@@ -8,6 +8,14 @@
 #import "LPWebQuery.h"
 #import "LPReflectUtils.h"
 #import "LPWebQuery.h"
+#import "LPInvoker.h"
+
+@interface UIScriptASTWith ()
+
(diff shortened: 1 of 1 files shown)


jmoody:
Will not merge.

PR #165, opened 2015-05-11
LPJSONSerializer can encode arbitrary objects

jmoody:
### Motivation

I was interacting with the Keychain API the other day and discovered an error on the server when trying to encode (to JSON) the output of one of the public Keychain API calls.  The result was a RuntimeError in the client.

The public Keychain API method returned an NSDictionary and one of the elements in that array was not a valid JSON object.

The error is occuring because of the the way `LPJSONUtils` is interacting with the LPJSONSerializer.

This was the common pattern:

```objective-c
 LPCJSONSerializer *s = [LPCJSONSerializer serializer];
  NSError *error = nil;
  NSData *d = [s serializeObject:obj error:&error];
  if (error) {
    NSLog(@"Unable to serialize object (%@), %@", error, [obj description]);
  }
  NSString *res = [[NSString alloc] initWithBytes:[d bytes] length:[d length]
                                         encoding:NSUTF8StringEncoding];
  return res;
```

1. The existence of `error` should not be used to branch; classes are allowed to scribble on errors passed by reference.  The LPJSONSerializer does not scribble, but Apple's APIs can and do.
2. `NSData` is `NULL` because `serializeObject:` could not serialize the object.  This causes `res` to be `"null"` which is not valid JSON.

The end result is JSON parse error.

```
JSON::ParserError: A JSON text must at least contain two octets!
```

@krukow On Friday afternoon, I was able to generate a server crash related to this issue, but now I cannot reproduce.  I will keep searching.

### DIFF
--- a/LPTestTarget/AppDelegate.h
+++ b/LPTestTarget/AppDelegate.h
@@ -1,7 +1,9 @@
 #import <UIKit/UIKit.h>
+@class NSManagedObjectContext;
 
 @interface AppDelegate : UIResponder <UIApplicationDelegate>
 
 @property (strong, nonatomic) UIWindow *window;
+@property (strong, nonatomic, readonly) NSManagedObjectContext *managedObjectContext;
 
 @end
--- a/LPTestTarget/AppDelegate.m
+++ b/LPTestTarget/AppDelegate.m
@@ -1,10 +1,20 @@
 #import "AppDelegate.h"
+#import <CoreData/CoreData.h>
+#import "LPCoreDataStack.h"
 
 @interface AppDelegate ()
 
 @end
 
 @implementation AppDelegate
+
+@synthesize managedObjectContext = _managedObjectContext;
+
+- (NSManagedObjectContext *) managedObjectContext {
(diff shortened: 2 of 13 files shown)


jmoody:
@krukow 

I used the same `ensure` concept I have been using in other places.  The serializer should always be able to return valid JSON.

The only time this is not true is if calling `description` raises an exception.   I added guards to catch exceptions when `description` is called on NSManagedObject.

jmoody:
@krukow Travis is still unstable.  XCTests are not starting and gherkin installation is failing.

jmoody:
@krukow In the end, I think I was incorrectly interpreting the signal I was getting from the client.  I saw:

```
JSON::ParserError: A JSON text must at least contain two octets!
```

and interpreted it as an application crash.  This occurred in the last step in the Scenario and the app immediately restarted.  This made me think it was crash.

This PR still has a bunch of merit.  It allows arbitrary objects to be encoded as JSON.  It also fixes the client side Keychain API which is broken without this fix.

PR #166, opened 2015-05-11
Don't include non-public LPHTTPServer.h in CalabashServer.h

ldiqual:
`LPHTTPServer.h` was included in `CalabashServer.h` even though it is not public. This caused a compilation error when including `CalabashServer.h` from the framework.

Tested using `make framework`.


jmoody:
@ldiqual You are correct.  That should not be included.  But you shouldn't need to include CalabashServer.h in your project.  We do, however want people to be able to extend server in their projects, so this patch makes sense.

PR #171, opened 2015-05-12
LPJSONUtils fix double quoting of strings in jsonifyObject:

jmoody:
### Motivation

**LPJSONSerializer can encode arbitrary objects** #165  made an unfortunate change in `LPJSONUtils jsonifyObject:` that in some cases strings to become double quoted in query results.

```
> query('tabBarButton', AL)
[
    [0] "\"Buttons\"",
    [1] "\"Text\"",
    [2] "\"Date\"",
    [3] "\"Scrolling Views\"",
    [4] "\"Sliders\""
]
```

The original implementation, before #165 looked like this:

```
 # A bunch of if statements to handle special cases.
 LPCJSONSerializer *s = [LPCJSONSerializer serializer];
 NSError *error = nil;
 if (![s serializeObject:object error:&error] || error) {
   return [object description];
}
return object;
```

That looked like the wrong behavior to me because we were returning a non-JSON object from the method.  It turns out, this behavior is necessary.  I think it is unsafe, but fixing it is beyond the scope of this PR.

FYI/BTW I would have caught this much sooner, but my Jenkins install is misbehaving (fixed).  It was the Xamarin Test Cloud that ultimately caught the error.

### DIFF
--- a/calabash/Classes/JSON/LPCJSONSerializer.h
+++ b/calabash/Classes/JSON/LPCJSONSerializer.h
@@ -57,6 +57,11 @@ extern NSString *const LPJSONSerializerDoesNotRespondToDescriptionFormatString;
 // @todo We should be bubbling exceptional cases up to the response and marking
 // it as producing invalid JSON.  Otherwise the user may never see that there
 // is a potential problem in their app.
+
+// @todo Remove this from the public API.  The stringByEnsuring* methods should
+//       be the only public interface.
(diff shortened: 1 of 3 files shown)


jmoody:
Jenkins test this please.

jmoody:
Jenkins test this please

jmoody:
Jenkins test this please.

jmoody:
Jenkins test this please.

jmoody:
Jenkins test this please.

PR #177, opened 2015-05-28
LPMapRoute: convert to ARC

jmoody:
### Motivation

~~The LPOrientationOperation requires no views.  LPMapRoute was skipping the call to orientation operation because of this.  I don't know what the downstream effects will be.~~

LPMapRoute needed some formatting love and it was a good candidate for converting to ARC.  In the process of converting to ARC, I realized that some of the methods were _only_ for the Frank plug-in.  I documented this and made the Frank calls much safer.

~~I am curious to see how this be476d0 change will affect the server.~~ It was bad.

### Update

+1 for CI and +1 for tests that help you understand the behavior of your software.

The _fix_ was a breaking change.  Now I can see that the behavior that prompted this fix was ultimately caused by a misuse of the server API by the client.
    
```
    ;; Bad
    map_route('/orientation', :orientation, :status_bar)
    
    ;; Good
    map_route(nil, :orientation, :status_bar)
```

This PR still has merit because it cleans up the LPMapRoute file and converts it to ARC.


jmoody:
Jenkins failed:  http://ci.endoftheworl.de:8080/job/Calabash-iOS-Server-PR/52/cucumber-html-reports/

PR #180, opened 2015-06-17
MapKit categories were part of no target

jmoody:
### Motivation

These MapKit categories were added by a contributor @nalbion on Aug 3, 2012.

They are not part of any target so they have just been dead code.

In my projects, I have used these categories to test MapKit views - they are extremely useful.  Even though I am voting for removal, I think these Categories are useful.  They are a great candidate for a Calabash plug-in.  When I say I have used the Categories, I mean that I manually added them to my project sources.

There are two problems.

1. They require that Calabash link the MapKit framework.  We could make the linking optional, but this has caused problems in the past for Enterprise customers.
2. In general, we eschew Categories on public classes because of the possibility of namespace collisions.  For example, if a client were using these Categories already, linking Calabash would cause a duplicate symbol link-time error.

So thanks to @nalbion for the contribution, but these should not be part of the Calabash server.


jmoody:
Rubygems is having an incident and Travis CI is failing.  Jenkins looks good.

PR #184, opened 2015-06-18
XCTests should pass when run against any target simulator or device.

jmoody:
### Motivation

Back in January, when I started to LPJSONUtils under test, I wrote some quick and dirty tests that only worked on iOS 8 iPhone 5s devices.  The command-line tool only runs the tests against this simulator.

This bad practice continued when I implemented the runtime loading of WKWebView support.

I was more than a little embarrassed when I tried to explain this to @sapieneptus in Aarhus.

There is still a some behavior that I don't understand yet with respect to how the center point of a rect is determined, but this PR moves me one step closer.

Unfortunately, some tests fail when in landscape mode because the center points do not match expectations.

### DIFF
--- a/XCTest/Tests/Utils/LPDeviceTest.m
+++ b/XCTest/Tests/Utils/LPDeviceTest.m
@@ -72,6 +72,43 @@ describe(@"LPDevice", ^{
     expect([device system]).notTo.beNil();
   });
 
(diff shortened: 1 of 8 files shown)


jmoody:
Travis is down. Jenkins and XTC look good.

PR #186, opened 2015-06-19
XCTests need a way to compare iOS versions

jmoody:
### Motivation

As part of **XCTests should pass when run against any target simulator or device.** #184 I added two methods to LPDevice:

* - (NSString *) iOSVersion
* - (BOOL) lessThanIOS8

I don't want to encourage the use of iOS version comparison in the server.  We do this in one place in `LPUIAUserPrefsChannel` where we must know the iOS version to understand the iOS Simulator environment.

I've replaced the two methods from LPDevice with inline C functions that can do equality and inequality comparisons.

Why do I need this?

The behavior of the server is different iOS versions.  My iOS 9 testing is demonstrating that I need to know the iOS version to write good tests.


jmoody:
Travis is down.  Merging.

PR #187, opened 2015-06-19
Refactor: rename LPLog => LPLogger

jmoody:
### Motivation

It looks like we will be able to use CocoaLumberjack.  I have a local branch that demonstrates this.  The LPLog class clashes with my strategy for including CocoaLumberjack.

Why do we want CocoaLumberjack in the server?  

* Our current CocoaHttpServer and CocoaAsyncSocket files are wildly out of date - last updated in 2011. Xcode 7 revealed dozens of deprecated calls in these files.  Both of these repos use CocoaLumberjack.  The upgrade path will be easier if we can use CocoaLumberjack.
* CocoaLumberjack is a _fantastic_ logging framework - we could do some very cool things with it.


jmoody:
Travis is down.  Merging.

PR #215, opened 2015-07-19
WIP Expand the LPBackdoorRoute capabilities

jmoody:
**FORCE PUSHED:** Tue Jun 21 12:53 CET

### Motivation

1. Catches the case where the 'arg' key on the `data` dictionary has a nil value.

2. Uses the LPInvoker to allow backdoors to have any signature of this form:

```
- (<ANY>) backdoorSelector(<ANY>) arg;
```

These are a low priority changes and will require some work to update the docs.

If this is rejected, I will merge the nil 'arg' value change.

Ideally, I would like to remove nearly all restrictions on method signatures.  Maybe disallow AppDelegate methods to be called... The docs work could be deferred until this is a reality.

At issue are these Scenarios:

```
 48   @expect_crash
 49   Scenario: Calling backdoor with void return type causes a crash
 50     And I call backdoor on a method whose return type is void
 51     Then the app should crash
 52 
 53   @expect_crash
 54   Scenario: Calling backdoor with primitive return type causes a crash
 55     And I call backdoor on a method that returns a primitive
 56     Then the app should crash
 57 
 58   @invalid_signature
 59   Scenario: Backdoors with primitive argument (NSUInteger) are unpredictable
 60     And I call backdoor on a method that takes an NSUInteger argument
 61     Then I won't get back that integer as a string
 62 
 63   @invalid_signature
 64   Scenario: Backdoors with primitive argument (BOOL) are unpredictable
 65     And I call backdoor on a method that takes a boolean argument
 66     Then I won't get back that boolean as a string
```

- [ ] @krukow needs review
- [x] @sapieneptus needs review

### TODO

- [ ] 1. Fix up the "should crash" Scenarios
- [ ] 2. Fix up the BOOL argument - LPInvoker is _too_ good at coercing encodings - it returns true/false for BOOL and not 0/1.
- [ ] 3. Remove guard against missing `:`
- [ ] 4. Change the reason/details returned when unrecognized selector is sent
- [ ] 5. Add tests for `@"__nil__"`

### DIFF
--- a/calabash/Classes/FranklyServer/Routes/LPBackdoorRoute.m
+++ b/calabash/Classes/FranklyServer/Routes/LPBackdoorRoute.m
@@ -21,7 +21,9 @@
   return [method isEqualToString:@"POST"];
 }
 
-- (NSDictionary *) JSONResponseForMethod:(NSString *) method URI:(NSString *) path data:(NSDictionary *) data {
+- (NSDictionary *) JSONResponseForMethod:(NSString *) method
+                                     URI:(NSString *) path
+                                    data:(NSDictionary *) data {
   NSString *originalSelStr = [data objectForKey:@"selector"];
   NSString *selectorName = originalSelStr;
   if (![originalSelStr hasSuffix:@":"]) {
--- a/calabash/Classes/FranklyServer/Routes/LPBackdoorRoute.m
+++ b/calabash/Classes/FranklyServer/Routes/LPBackdoorRoute.m
@@ -11,6 +11,9 @@
 
 #import "LPBackdoorRoute.h"
 #import "LPCocoaLumberjack.h"
+#import "LPInvoker.h"
+#import "LPInvocationResult.h"
+#import "LPInvocationError.h"
 
 @implementation LPBackdoorRoute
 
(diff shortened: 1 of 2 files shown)


calabash/Classes/FranklyServer/Routes/LPBackdoorRoute.m
```diff
@@ -29,40 +34,41 @@ - (NSDictionary *) JSONResponseForMethod:(NSString *) method URI:(NSString *) pa
     selectorName = [selectorName stringByAppendingString:@":"];
   }
 
+  id argument = [data objectForKey:@"arg"];
```
sapieneptus: Do we not like `data[@"arg"]` syntax?

jmoody: We _love_ it.

Change it!

calabash/Classes/FranklyServer/Routes/LPBackdoorRoute.m
```diff
@@ -29,40 +34,41 @@ - (NSDictionary *) JSONResponseForMethod:(NSString *) method URI:(NSString *) pa
...
+  if (!argument) {
+    LPLogError(@"Expected data dictionary to contain an 'arg' key");
```
sapieneptus: Is it really a problem is the user doesn't supply an argument? Objective-C was designed to be robust to `nil` pointer operations, and it's acceptable in general to pass `nil` as a method param. Plus, idiomatically, it's odd to write methods like:

`- (void)changeAppState:(id)ignoreMe { /**/ } `

sapieneptus: The exception of course being methods with `__non_null` parameters, but if the user is trying to call those and is passing nil, they need to have a convo with their devs.

jmoody: If we are going to expand the method signatures, then +1 for not requiring an argument _and_ for passing multiple arguments.

I _think_ however that the client should always provide an arg key (or "arguments") key.  But I am not picky.

jmoody: I was hoping this PR could be a non-breaking, incremental change.

sapieneptus: Meta-language manipulations is one of my favorite things ever. I'm _sure_ there's a way to do it in a way that's non-breaking. 

As soon as we hit the `NSSelectorFromString()` call, the number and type of arguments shouldn't matter (at least insofar as the ObjcRuntime is concerned). Looks like it's more a question of how `LPInvoker` can handle `nil`s.

jmoody: If you pass an empty array to:

```
LPInvoker invokeSelector:withTarget:arguments:
```

_and_ the selector takes no argument, all will be well.

The argument count must match the number of arguments in the selector.

LPInvoker doesn't handle `[NSNull null]` as nil.

sapieneptus: Hmm... this creates an ambiguous case. Treating all `[NSNull null]` as `nil` would mean that users can't invoke using literal `null`, which could be problematic if the user was trying to insert values into a dict or something. 

But treating all `[NSNull null]` as proper `null` will lead to NPEs if the user is expecting it to be treated as `nil`. 

I can imagine a few ways around this, but they're awkward and require more changes than we'd probably want to deal with now. 
E.g.:

```
//User wants to invoke a selector with sig:
// - (id)foo:(id)foo bar:(id)bar baz:(id)baz;
{
    @"method" : @"foo:bar:baz:",
    @"arguments" : @{
        @"foo:" : @"a value",
        @"baz" : [NSNull null]
    }
}

//The method would be invoked with:
[target foo:@"a value" bar:nil baz:[NSNull null]];
```

sapieneptus: Hey... wait... 

What's this [`__self__` keyword](https://github.com/calabash/calabash-ios-server/blob/develop/calabash/Classes/Invoker/LPInvoker.m#L651)? Is this an Obj-C thing or something we're introducing?

If it's the latter, could we not have a `@"__nil__"` ?

jmoody: re: NSNull - You are correct.  I started working on that, but didn't stress about it because there are other things to focus on.  

@re:  `__self__` - https://github.com/calabash/calabash-ios-server/blob/develop/XCTest/Tests/Invoker/LPInvokerSelectorWithArgumentsTest.m#L348

We could!  Pretty good idea!

sapieneptus:
It appears I broke everything...

PR #220, opened 2015-07-27
Expand backdoor route to handle arbitrary method signatures

sapieneptus:
**FORCE PUSHED** rebased Tue 20 Oct 15:04

- Modified `LPBackdoorRoute` to be able to handle `args` key which should be an array of args. This allows invocation of methods with multiple params. 
- Modified `LPInvoker` to accept `@"__nil__"` as nil, thereby allowing clients to pass `nil` as a param to a backdoor. 
- Since I _believe_ a void return type is already handled, this should allow for backdoor methods of arbitrary sigs.

### DIFF
--- a/calabash/Classes/FranklyServer/Routes/LPBackdoorRoute.m
+++ b/calabash/Classes/FranklyServer/Routes/LPBackdoorRoute.m
@@ -16,8 +16,8 @@
 #import "LPInvocationResult.h"
 #import "LPInvocationError.h"
 
-const static NSString *ARG_KEY = @"arg";  /* for backwards compatibility */
-const static NSString *ARGUMENTS_KEY = @"arguments";
+static NSString *const ARG_KEY = @"arg";  /* for backwards compatibility */
+static NSString *const ARGUMENTS_KEY = @"arguments";
 
 @implementation LPBackdoorRoute
 
--- a/calabash/Classes/FranklyServer/Routes/LPBackdoorRoute.m
+++ b/calabash/Classes/FranklyServer/Routes/LPBackdoorRoute.m
@@ -120,7 +120,7 @@ static NSString *const ARGUMENTS_KEY = @"arguments";
 
     NSString *details = [lines componentsJoinedByString:@"\n"];
 
-    NSString *reason = [NSString stringWithFormat:@"The backdoor: '%@' is undefined.",
+    NSString *reason = [NSString stringWithFormat:@"The backdoor: '%@' is undefined",
                         selectorName];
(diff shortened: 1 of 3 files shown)


calabash/Classes/FranklyServer/Routes/LPBackdoorRoute.m
```diff
@@ -11,65 +11,65 @@
...
+    return [self failureWithReason:@"Missing selector name."
+                           details:[NSString stringWithFormat:@"Expected 'arg' OR 'args' key in data, not both. Data: '%@'", data]];
+  } else if (!(data[@"arg"] || data[@"args"])) {
+    LPLogError(@"Expected data dictionary to contain an 'arg' or 'args' key.\nData = %@", data);
+    return [self failureWithReason:[NSString stringWithFormat:@"Missing argument(s) for selector: '%@'",
+                                    selectorName]
+                           details:[NSString stringWithFormat:@"Expected backdoor selector '%@' to have an argument(s), but found no 'arg' or 'args' key in data '%@'", selectorName, data]];
   }
+  
+  id arguments = data[@"arg"] ? @[data[@"arg"]] : data[@"args"];
```
jmoody: Keep the @"arg" for backwards compatibility, but I think the key should be "arguments".

sapieneptus: Yeah, was thinking that as well.

jmoody:
@sapieneptus I rebased!  I recommend dropping your local branch and fetching a fresh copy?

jmoody:
**LPInvoker handles __self__ and __nil__ tokens in arguments** #270 replaces a commit in this PR.

jmoody:
Updated with some unit tests for the unhappy paths: 

* missing or duplicate keys
* unknown selector

I updated the cucumbers in [backdoor.feature](https://github.com/calabash/ios-smoke-test-app/blob/master/CalSmokeApp/features/backdoor.feature) in a local branch to reflect the new capabilities.

The client will need a fix that looks like this:

```ruby
def backdoor(selector, *arguments)
        json = {
              :selector => selector,
              :arguments => arguments
        }
        res = http({:method => :post, :path => 'backdoor'}, json)
        res = JSON.parse(res)
        if res['outcome'] != 'SUCCESS'
          screenshot_and_raise "backdoor #{json} failed because:\n\n#{res['reason']}\n#{res['details']}"
        end
        res['result']
      end
```

jmoody:
@sapieneptus I think we can merge this.

jmoody:
Decided to merge because it turns out we will be releasing 0.17.  This is potentially a breaking change.

PR #221, opened 2015-07-29
new route to lookup the current keyboard's l10n

svevang:
This PR sets up a new calabase server route that returns a localization code for the open keyboard.

In subsequent pull requests, I'll lay out some facilities for using this localization code in calabash-ios to look up the name of various buttons and keys via UIKit's localizations present in xcode.


jmoody:
@svevang Can you test my changes against the Calabash iOS and run-loop PRs?

PR #222, opened 2015-08-03
Ensure all ThirdParty library symbols are namespaced

jmoody:
### Motivation

While trying to build the WordPress example, I discovered I missed a CocoaLumberjack global function.

```
doesAppRunInBackground
```

Fixed that issue.

Added CocoaLumberjack, CocoaAsyncSocket, and CocoaHTTPServer to the LPTestTarget which is the app that is run with the XCTests.  If there is duplicate symbol, we'll know because the XCTests will fail.


jmoody:
@krukow @sapieneptus  I missed one. :(  But this PR ensures that I haven't missed any others.

PR #241, opened 2015-09-14
LPUIAUserPrefsChannel: use NSUserDefaults

sapieneptus:
I don't have any facts here, but what I found was that if the `.plist` was in binary form, it seemed to work. If it was in plain text, it did not. `[[NSUserDefaults standardDefaults] setObject:forKey]` seems to write to that file synchronously as far as I can tell, but will do more testing to confirm (as well as backwards compat w/ 8.x sims).

Note: appears to work on 8.3 simulator, didn't notice any issues with the `index` being out of sync or anything. It _appears_ to be a little faster in 8.3 than 9.0.


jmoody:
@sapieneptus @krukow I am closing this in favor of a PR that preserves the iOS < 9.0 behavior.

Issue #266, opened 2015-10-16
Migrate project to ARC so all new files are ARC by default


jmoody:
The policy is: 

1. All new files should be ARC
2. Every time you touch a file, you should evaluate it for conversion to ARC.

Issue #236, opened 2015-09-11
Xcode 7: preferences strategy is not working for iOS >= 9.0 simulators


jmoody:
Check if the server is detecting the plist correctly.

jmoody:
@sapieneptus 

We should verify that UIAutomation API is working in the Instruments.app.

jmoody:
- [x] Verified that UIAutomation API is working in the Instruments.app
- [x] Discovered that the validateRequestWritten: is not finding the UIA request even though the plist on disk has been written.

jmoody:
Verified that :preferences works for iOS Simulators < 9.0

jmoody:
Testing a solution here **Modernize and Calabashify** [#2](https://github.com/JaniJegoroff/calabash-ios-example-issue-585/pull/2)

So far so good.  I can't reproduce the problem.

jmoody:
- [ ] Setup some CI for this issue - waiting on @ncouraud to configure public URL for new Calabash Jenkins instance

PR #267, opened 2015-10-16
Fix :preferences strategy for iOS 9

jmoody:
### Motivation

**Xcode 7: preferences strategy is not working for iOS >= 9.0 simulators** #236

The gist of this pull request is to branch on iOS 9 and interact directly with NSUserDefaults rather than do IO directly on the simulator plist.

Could not reproduce the timing issue [#585](https://github.com/calabash/calabash-ios/issues/585) with this https://github.com/JaniJegoroff/calabash-ios-example-issue-585/pull/2

Progress on:

* **Migrate project to ARC so all new files are ARC by default** #266
* **Replace NSLog calls with LPLogDebug/Info** #232

As mentioned by @krukow  in [this issue](https://github.com/xamarin/test-cloud-frontend/issues/1710) we should release to allow users like @JaniJegoroff, @ark-konopacki, @nfrydenholm to test. 

The contents of https://github.com/xamarin/test-cloud-frontend/issues/1710 should be public - the discussion there is extremely interesting and it is a drag to keep referencing it.


nfrydenholm:
Does this mean there will be a pre release with these fixes, so we can try the preferences strategy?

We have started the progress of upgrading to SDK 9 (and latest Calabash version), and can now run our "migrate-branch" on an isolated Xcode7/SDK9 buildnode. On the first couple of runs, I have noticed a significant increase in time - will the preferences strategy help with that? And should we overall expect slower tests, as the simulator seems a bit slow in starting :worried:

jmoody:
> slower tests

:preferences is faster than :host.

> simulator is slow

iOS 9 Simulators have "booting" state.  If we try to install/uninstall or launch an app with instruments, we will get failures.  In the recent releases of run-loop, we've been _waiting_ for a stable simulator state.  I think I can relax the stable state heuristics in the next version of run-loop.

jmoody:
> Does this mean there will be a pre release with these fixes, so we can try the preferences strategy?

I think we will just cut a release next week and keep the default strategy as :host.  Savvy users can opt-in to :preferences.

PR #269, opened 2015-10-19
Reflection route

jmoody:
@sapieneptus Can you have a look?  Useful?

### DIFF
--- a/calabash/Classes/FranklyServer/Routes/LPReflectionRoute.h
+++ b/calabash/Classes/FranklyServer/Routes/LPReflectionRoute.h
@@ -1,0 +1,6 @@
+#import <Foundation/Foundation.h>
+#import "LPRoute.h"
+
+@interface LPReflectionRoute : NSObject<LPRoute>
+
+@end
--- a/calabash/Classes/FranklyServer/Routes/LPReflectionRoute.m
+++ b/calabash/Classes/FranklyServer/Routes/LPReflectionRoute.m
@@ -8,6 +8,7 @@
 @interface LPReflectionRoute ()
 
 - (NSArray *) libraryNames;
+- (NSArray *) classNames;
 
 @end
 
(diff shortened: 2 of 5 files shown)


calabash/Classes/FranklyServer/Routes/LPReflectionRoute.h
```diff
@@ -0,0 +1,6 @@
+#import <Foundation/Foundation.h>
+#import "LPRoute.h"
+
+@interface LPReflectionRoute : NSObject<LPRoute>
```
sapieneptus: Just out of curiosity, is there an advantage of using a `@protocol` vs inheritance? I know this is how all of the routes are made, strictly curious.

calabash/Classes/FranklyServer/Routes/LPReflectionRoute.m
```diff
@@ -0,0 +1,67 @@
...
+           @"classes" : [self classNames]
+           };
+}
+
+- (NSArray *) libraryNames {
+  unsigned int number = 0;
+
+  const char **names = objc_copyImageNames(&number);
+
+  NSMutableArray *array = [NSMutableArray arrayWithCapacity:number];
```
sapieneptus: It might be possible to modularize this function into a generic `- (NSArray *)names( rawGeneratorBlock )` so that you could use the same function to get both library names and class names, but it's probably more readable this way.

sapieneptus:
@jmoody Is there a corresponding PR on the client?

sapieneptus:
And yes, `useful++ `

calabash/Classes/FranklyServer/Routes/LPReflectionRoute.h
jmoody: I can't think of an advantage especially since none of the methods in the protocol are required.

calabash/Classes/FranklyServer/Routes/LPReflectionRoute.m
jmoody: Yeah, I was thinking 3 would be a magic number for extracting a generic method.

jmoody:
> PR on client

No.  I haven't decide whether or not to support this in the client.  Once it is added there it will be hard to remove.  If you feel strongly about having this in the client, please let me know.

```
 JSON.parse(http({:method => :get, :path => 'reflection'}, {}))
```

> useful ++

Good good.

Issue #279, opened 2015-10-28
Xcode 7.1: `make framework` does not find arm library on every system


jmoody:
Reported by @stopiccot

Changing the ARM_LIBRARY_XC7 fixes the problem:

```
"${ARM_BUILD_DIR}/Build/Intermediates/ArchiveIntermediates/${XC_TARGET}/IntermediateBuildFilesPath/UninstalledProducts/${LIBRARY_NAME}"`
"${ARM_BUILD_DIR}/Build/Intermediates/ArchiveIntermediates/${XC_TARGET}/IntermediateBuildFilesPath/UninstalledProducts/iphoneos/${LIBRARY_NAME}"`
```

jmoody:
Discovered the problem - the paths are indeed incorrect on the master branch.

Fixed on the develop branch.

Thanks @stopiccot!

jmoody:
Well, now I have no idea what is going on.

There are definitely problems with the build system.

stopiccot:
Obviously I don't see the whole picture and may be missing something, but I strongly advice switching to framework building by Xcode. This way you don't need to guess where yet-another-new-Xcode-in-which-Apple-changed-everything-again stores files.

jmoody:
It has been proposed and I am interested.   I did some framework exploration recently and found that there were some downsides.

 This method has served us pretty well.  And this is the first time we've encountered something like this.

PR #280, opened 2015-10-29
Add 'shake' route to mimic shake of device on iOS9+ devices

tommeier:
As discussed in https://github.com/calabash/calabash-ios/issues/873

iOS9+ devices have no ability to mimic shake via UIAutomation, I've filed radar://22796974 but no response.

In the meantime here is an implementation to mimic the shake. I've tested on simulator, but not on device yet. I haven't been able to get the tests running on this suite, I assume I'm missing a dependency or something?

screen shot 2015-10-29 at 8 34 02 pm

This should only be used for iOS9+ devices as `target.shake()` continues to work for iOS 8.4 and below (even in XCode 7.1).

### DIFF
--- a/cucumber/features/steps/alert_steps.rb
+++ b/cucumber/features/steps/alert_steps.rb
@@ -1,0 +1,83 @@
+
+def alert_exists?(alert_title=nil)
+  if alert_title.nil?
+    res = uia('uia.alert() != null')
+  else
+    if ios6?
+      res = uia("uia.alert().staticTexts()['#{alert_title}'].label()")
+    else
+      res = uia("uia.alert().name() == '#{alert_title}'")
+    end
+  end
+
+  if res['status'] == 'success'
+    res['value']
+  else
+    false
+  end
+end
+
+def wait_for_alert
+  timeout = 4
+  message = "Waited #{timeout} seconds for an alert to appear"
+  options = {timeout: timeout, timeout_message: message}
+
+  wait_for(options) do
+    alert_exists?
+  end
+end
+
+def alert_view_query_str
+  if ios8? || ios9?
+    "view:'_UIAlertControllerView'"
+  elsif ios7?
+    "view:'_UIModalItemAlertContentView'"
+  else
+    'UIAlertView'
+  end
+end
+
+def alert_button_views
+  wait_for_alert
+
+  if ios8? || ios9?
+    query = "view:'_UIAlertControllerActionView'"
+  elsif ios7?
+    query = "view:'_UIModalItemAlertContentView' descendant UITableView descendant label"
+  else
+    query = 'UIAlertView descendant button'
+  end
+  query(query)
+end
+
+def alert_button_titles
+  alert_button_views.map { |res| res['label'] }.compact
+end
+
+def all_alert_labels
+  wait_for_alert
+  query = "#{alert_view_query_str} descendant label"
+  query(query)
+end
+
+def non_alert_button_views
+  button_titles = alert_button_titles()
+  all_labels = all_alert_labels()
+  all_labels.select do |res|
+    !button_titles.include?(res['label']) &&
+          res['label'] != nil
+  end
+end
+
+def alert_message
+  with_max_y = non_alert_button_views.max_by do |res|
+    res['rect']['y']
+  end
+
+  with_max_y['label']
+end
+
+Then(/^I (should )?see the "([^"]*)" alert$/) do |_should, message|
+  wait_for_alert
+  expect(alert_message).to be == message
+end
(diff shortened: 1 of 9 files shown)


jmoody:
Woot!

My best guess that you did not add the -fobjc-arc flag to the relevant files.

You can verify by running without xcpretty:

```
$ make clean
$ XCPRETTY=0 make framework
```

The command line scripts treat warnings as errors.  We do this because everyone forgets to add the -fobjc-arc flag and this warning may or may not appear in Xcode - or might be ignored.

The server is, by default, a non-ARC project which is a drag.  

I made a short video to show you how to add a new route to the project.

[screencast.com ~3:30](http://www.screencast.com/t/eol5CfDwXZzi)

1. Add the new route to following targets:
  * XCTest
  * calabash
  * calabash-plugin-for-frank
  * calabash-dylib
2. Remove all Copyright info comments; these are useless and resist refactoring.  If you want credit, you can add your name and a url.
3. Add the -fobjc-arc flag to the new route files
4. There are no Project-level headers for this project.  Remove them from the targets' headers.
5. Add unit test file, but only to the XCTest target.
6. Ditto copyright
7. Ditto -fobjc-arc flag

And thank you thank you thank you!

jmoody:
@tommeier 

There is also this problem, which I am hunting down:

**Xcode 7: `make framework` does not find arm library on every system** #279

Until I have a fix for that make framework will probably fail.

tommeier:
@jmoody thanks, added the missing flags to the file, noticed there were quite a few others without it too, but left them alone. Now I hit the Xcode 7 issue (at a guess): 

```
can't get real path for source 'build/framework/arm/Build/Intermediates/UninstalledProducts/libcalabash.a'
ERROR: Could not copy:
ERROR:   source: build/framework/arm/Build/Intermediates/UninstalledProducts/libcalabash.a
ERROR:   target: Products/framework/arm/libcalabash.a
ERROR: The source file does not exist
ERROR: Did a previous xcodebuild step fail?
ERROR: Exiting 1
```

Will see if CI has better luck this time.

tommeier:
@jmoody CI looking happy now, how do I add a smoke test? where are the CI smoke test files located?

jmoody:
> noticed there were quite a few others without it too, 

There is a mix of ARC and non-ARC files in this project.

> Now I hit the Xcode 7 issue 

Please rebase against develop; I merged the fix.  I now believe it is an Xcode 7.0.1 vs Xcode 7.1 problem.

> how do I add a smoke test?

That is a very good question.  Until last week, I would have said, write some Cucumbers in this project:

https://github.com/calabash/ios-smoke-test-app

but the role of the project recently changed and we have to be more careful that the Cucumbers there are always passing.

You and I are in convenient timezones I think - I will get back to you about this.

And thank you very very much!

jmoody:
@tommeier 

Standup cucumber tests against LPTestTarget #283

There is now a cucumber/ directory and the scaffolding for writing cucumbers against the LPTestTarget.

Don't be afraid to rebase!

tommeier:
@jmoody fixed, rebased and squished, can now run full suite and tests. 

I've also added a cucumber test for checking the shake command route, and in addition added an optional 'duration' parameter, in case users check that the 'shake' has occured for 'x' seconds (common use case to not trigger excessively).

jmoody:
@tommeier 

Excellent!  Nice work!  Thanks for your patience!

cucumber/features/steps/alert_steps.rb
```diff
@@ -0,0 +1,83 @@
...
+end
+
+def alert_message
+  with_max_y = non_alert_button_views.max_by do |res|
+    res['rect']['y']
+  end
+
+  with_max_y['label']
+end
+
```
jmoody: @tommeier 

When defining helper methods, it is better to wrap them in a module - otherwise the methods are added to the Kernel namespace.

```ruby
module LPTestTarget
  module Alert

  end
end

World(LPTestTarget::Alert)
```

I am adding this change, so there is no need for you to do anything further.

Thanks!

tommeier:
No worries :)

tommeier:
@jmoody FYI this is the response from Apple:

```
This is a follow-up to Bug ID# 22796974.   

Engineering has provided the following feedback regarding this issue: 

The Automation Instrument is deprecated and is no longer supported. Please migrate your testing to the UI Testing feature in Xcode 7.

We consider this issue closed.  If you have any questions or concerns regarding this issue, please update your report directly here: http://bugreport.apple.com/.

Thank you for taking the time to notify us of this issue.

Best Regards,

Developer Bug Reporting Team 
Apple Worldwide Developer Relations
```

Not sure how this will affect things going forward, but I imagine it will mean a lot more workarounds.

PR #288, opened 2015-11-11
LPShakeRoute needs to respond to POST and include :results key in JSON response

jmoody:
### Motivation

While implementing the shake client and working the LPSuspendAppRoute, I remembered two things:

* SUCCESS responses must include "results" key in the JSON
* Routes that include parameters need to respond to POST.

Thanks again to @tommeier

### DIFF
--- a/LPTestTarget/FirstViewController.m
+++ b/LPTestTarget/FirstViewController.m
@@ -2,6 +2,8 @@
 
 @interface FirstViewController ()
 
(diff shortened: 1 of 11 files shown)


jmoody:
Also fixes the iOS 7 failure [on the XTC](https://testcloud.xamarin.com/test/sh-calaba-lptesttarget_2951dc6a-5a65-432c-b794-41ff77ad9536/).

PR #292, opened 2015-11-20
LPTestTarget: force a new app SHA for every build

jmoody:
The sha of the app does not change when a new calabash.framework is
linked.  When testing new server change against the iOS simulator
we need the new .app with the new server changes to be installed
before testing.  This change forces the app to have a new sha by
modifying LPTestTarget/change-sha.plist with the date of the build.

The change-sha.plist is reverted (git checkout) after the build so
we don't have to track changes in this plist.


jmoody:
Cucumbers are still failing for the iPad Air.

PR #293, opened 2015-11-20
Standup Jenkins in MacMini Colo: part 2

jmoody:
### Motivation

1. The cucumber tests are failing, but I can't easily tell which ones - using Slowhand formatter
2. Submit an XTC job


jmoody:
Woot!  Jenkins!  Done me proud!

Issue #303, opened 2015-12-08
Xcode 7.2: builds simulator dylib to Debug-iphoneos directory


jmoody:
instead of Debug-iphonesimulator directory:

```
$ xcrun lipo -info build/dylib/sim/Build/Products/Debug-iphoneos/calabash-dylib.dylib 
Architectures in the fat file: build/dylib/sim/Build/Products/Debug-iphoneos/calabash-dylib.dylib are: i386 x86_64 
```

jmoody:
```
TARGET_BUILD_DIR="${BUILD_PRODUCTS_DIR}"                                                  
DWARF_DSYM_FOLDER_PATH="${BUILD_PRODUCTS_DIR}" 
```

ohadschn:
Are these settings the workaround for this issue or simply your current setting with which it reproduces? Did you encounter any issues with the fact that it builds to the wrong folder or is it just a cosmetic oversight?

jmoody:
100% confirmed that simulator targets are built to -iphoneos vs. -iphonesimulator.  Also happens when building non-library targets (like apps for the simulator).

The scripts fail because they expect binaries to be in a particular place.

> Are these settings the workaround for this issue or simply your current configuration with which it reproduces? 

I am debugging another (related) issue and found if I set those variables, I can coerce xcodebuild to produce the expected directories.

I will update the list in a moment.

jmoody:
> Did you encounter any issues with the fact that it builds to the wrong folder or is it just a cosmetic oversight?

I don't know yet.  I suspect there will be a problem because the sim and arm Intermediate directory will be the same, but I cannot confirm.

ohadschn:
Thanks for the speedy response! When you say "the scripts fail" do you mean calabash scripts or Apple scripts?

jmoody:
The xcodebuild scripts.

The Calabash 0.17.0 libraries are fine; we built them with Xcode 7.1.

ohadschn:
Oh, so what xcodebuild command exactly fails because of that (and with what error)? We've seen some failures with 7.2 and I'm wondering if they're related (I had to explicitly specify the VALID_ARCHS  flag).

jmoody:
> We've seen some failures with 7.2 and I'm wondering if they're related

What are you building?

```
$ make dylibs
⚠️  ld: directory not found for option '-L/Users/moody/git/calabash/calabash-ios-server/build/dylib/sim/Build/Products/Debug-iphoneos'
[...]

Issue #302, opened 2015-12-08
Xcode 7.2: builds frank simulator library to Debug-iphoneos directory


jmoody:
and not to the Debug-iphonesimulator directory:

```
$ xcrun lipo -info build/frank-plugin/sim/Build/Products/Debug-iphoneos/libcalabash-plugin-for-frank.a 
Architectures in the fat file: build/frank-plugin/sim/Build/Products/Debug-iphoneos/libcalabash-plugin-for-frank.a are: i386 x86_64 
```

jmoody:
```
TARGET_BUILD_DIR="${BUILD_PRODUCTS_DIR}"                                                  
DWARF_DSYM_FOLDER_PATH="${BUILD_PRODUCTS_DIR}" 
```

PR #305, opened 2015-12-21
Update build scripts for Xcode 7.2

jmoody:
### Motivation

Resolves:

* Xcode 7.2: builds frank simulator library to Debug-iphoneos directory #302
* Xcode 7.2: builds simulator dylib to Debug-iphoneos directory #303

The problem is Xcode 7.2 requires:

```
EFFECTIVE_PLATFORM_NAME="-iphonesimulator"
```

### To Review

Reviewer must have Xcode 7.2

```
$ git fetch
$ git clone -t origin/feature/update-build-scripts-for-xcode-72
$ bundle install
$ make frank
$ make dylib
$ make app-cal
```


bin/make/frank-plugin.sh
```diff
@@ -95,6 +101,7 @@ xcrun xcodebuild build \
   ARCHS="i386 x86_64" \
   VALID_ARCHS="i386 x86_64" \
   ONLY_ACTIVE_ARCH=NO \
+  EFFECTIVE_PLATFORM_NAME="-iphonesimulator" \
```
sapieneptus: I don't really have enough context to ask good questions here, so please bear with me- but why is this line necessary ?

jmoody: Starting in Xcode 7.2, simulator binaries (apps and libraries) are being built to `Debug-iphoneos` and not to `Debug-iphonesimulator`.   After exploring a bunch of other build settings, I found that this was the missing argument.

jmoody: In case it was not clear, we could not build the dylibs or frank libraries with Xcode 7.2.

sapieneptus: Got it - How did you figure out that this was the missing arg? Seems like a very non-obvious thing to have caused that result.

jmoody: It was not the first thing I tried.

I started by going through the Xcode build logs from Xcode looking at what variables were available to xcodebuild.  I came up with this brute force approach which forced Xcode to build to very specific directories.


```
OBJECT_ROOT_DIR="${XC_BUILD_DIR}/Build/Intermediates/${XC_CONFIG}-iphonesimulator"

xcodebuild
    ...
    OBJROOT="${OBJECT_ROOT_DIR}" \
    SHARED_DERIVED_FILE_DIR="${BUILD_PRODUCTS_DIR}" \
    CONFIGURATION_BUILD_DIR="${BUILD_PRODUCTS_DIR}" \
    BUILT_PRODUCTS_DIR="${BUILD_PRODUCTS_DIR}" \
    TARGET_BUILD_DIR="${BUILD_PRODUCTS_DIR}" \
    DWARF_DSYM_FOLDER_PATH="${BUILD_PRODUCTS_DIR}" \
```

This works, but it is a lot of effort.  I noticed the `EFFECTIVE_PLATFORM_NAME` variable early on but thought, "That can't be the problem...".   Turns out that it was. :8ball:

PR #310, opened 2016-01-04
Fix JavaScript insertion into LPWebQuery.h

jmoody:
### Motivation

The `make webquery-header` rule was only building the minified JavaScript, it was not inserting the new JS into the LPWebQuery.h file.


sapieneptus:
Looks ok to me - IMO we ought to do something about the references to `enter_text.js`, (which is part of `build.sh`) but that should be a separate PR.

sapieneptus:
Actually, @jmoody, `enter_text.js` is still in use in `LPSetTextOperation`. So we should definitely extend this PR to cover that one as well (it's already minified in `build.sh`, so it's just a matter of inserting it into the header).

Makefile
```diff
@@ -21,8 +21,8 @@ dylibs:
 	# See https://github.com/calabash/calabash-codesign for details.
 	bin/make/dylibs.sh 337976ad9ace375ac06cd8fea2edb0c7276dec2a72d005ca5559a8bbf09c8841
 
-webquery_headers:
-	bundle exec bin/make/insert-js-into-webquery-headers.rb
+webquery_header:
```
sapieneptus: @jmoody In light of discovering that `set_text.js` is still used, I think it should actually be 'webquery_headers' :)

jmoody:
> set_text.js

As far as I can tell, this was never part of the original calabash-js / server integration.  Put another way, I don't see that set_text.js was ever inserted into LPSetTextOperation.h.

PR #311, opened 2016-01-05
Make: add xctests rule from running XCTest

jmoody:
### Motivation

More than once I have instructed users to run `make xtc` instead of `make xct`.

The xct remains, but is now an alias for xctests.

FYI/BTW, 'make xctest' was my first choice, but this conflicts with
make's behavior:  if a directory with same name as the rule exists
the rule will exit "xctest is up to date".  The XCTests are in a
directory called 'XCTest'.


sapieneptus:
Approve :)

PR #313, opened 2016-01-06
Xcode: remove LPConstants.h from Project Headers

jmoody:
**REBASED** Wed Jan 6 17:55 CET

### Motivation

LPConstants.h should not be in the Project Headers.

We can discuss if it should be in the Public Headers, but think not.

### DIFF
--- a/calabash.xcodeproj/project.pbxproj
+++ b/calabash.xcodeproj/project.pbxproj
@@ -7,9 +7,6 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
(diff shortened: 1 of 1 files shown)


jmoody:
Travis is failing because of a codesigning issue when building the dylibs.

I will investigate in another PR, but all other tests passed.

sapieneptus:
@jmoody I guess a rebase is in order...

jmoody:
**REBASED** Wed Jan 6 17:55 CET

PR #314, opened 2016-01-06
LPConstants: needs to be compiled with ARC

jmoody:
### Motivation

This file does not technically need to be compiled with ARC, but we should mark it as needing ARC.

The issue is that we will forget that this file is _not_ compiled with ARC and then we add a leak without even knowing it.  This has happened more times than I can remember.


sapieneptus:
Good call

sapieneptus:
CI is failing for an unrelated reason. This file still compiles so this change can't be causing any problems.

PR #315, opened 2016-01-06
LPConstants: prefer NSString *const pattern

jmoody:
### Motivation

I am hazy on the details, but I believe this is the correct way to declare extern'd constants.

This is non-blocking.  Feel free to close if you are not convinced or can argue against.

### DIFF
--- a/calabash/Classes/FranklyServer/Operations/LPConstants.h
+++ b/calabash/Classes/FranklyServer/Operations/LPConstants.h
@@ -8,11 +8,11 @@
 
 #import <Foundation/Foundation.h>
 
-extern NSString *NODE_NAME_KEY;
-extern NSString *IFRAME_KEY;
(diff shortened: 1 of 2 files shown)


sapieneptus:
A few notes:

originally, I had:
```Objective-C
// .h
extern NSString *NODE_NAME_KEY;
// .m
const NSString *NODE_NAME_KEY = @"nodeName";
``

I believe what happens is that if you try to assign `NODE_NAME_KEY` a new value, it should crash at runtime. Your approach is slightly better in this regard since it would halt you at compile time if `const` is in the header. 

Regarding `const NSString *` vs `NSString *const` in the .m file, is this just style preference? I'm unaware of any difference in effect between the two.

jmoody:
Style only I think.  We discussed off channel.  No need to reply here.

PR #316, opened 2016-01-06
LPWebQuery: make LP_QUERY_JS a const

jmoody:
### Motivation

This needs some attention I think.

When I change LP_QUERY_JS to a `*const* I get this compiler warning:

```
Classes/WebViewQuery/LPWebQuery.m:55:45: 
More '%' conversions than data arguments
```

I see four `%@` string and only three in the `stringWithFormat:` call.

Blocking.


sapieneptus:
@jmoody You're right - I guess the compiler didn't know because the value of `LP_QUERY_JS` is not statically analyzable without a `const` modifier. I've pushed a commit that should fix the missing arg.

jmoody:
@sapieneptus This cannot be rebased, so I am going to merge as-is.

Issue #320, opened 2016-01-11
build scripts should run the static analyzer


jmoody:
Today I found 2 issues:

* memory leak in a PR
* a bug in LPScrollOperation (introduced by me) [#980](https://github.com/calabash/calabash-ios/issues/980)

The analyzer is particularly important because this is a mixed ARC and retain/release project.

We have one trivial and one non-trivial problem.

![2016-01-11_11-49-32](https://cloud.githubusercontent.com/assets/466104/12231654/7ca28ff8-b859-11e5-998f-e6844fa4a666.png)

I believe that we could make the CalabashServer a singleton to resolve this issue.

PR #321, opened 2016-01-11
Fix LPScrollOperation for WebView

jmoody:
### Motivation

The CalWebApp has been failing because scrolling stopped working.

I introduced a bug in October 2015 that caused this.

Fixes **Scrolling on UIWebView and WKWebView is not work working in iOS 9** [#980](https://github.com/calabash/calabash-ios/issues/980)

Discovered this by running the static analyzer.

* build scripts should run the static analyzer #320

### DIFF
--- a/calabash/Classes/FranklyServer/Operations/LPScrollOperation.m
+++ b/calabash/Classes/FranklyServer/Operations/LPScrollOperation.m
@@ -13,6 +13,7 @@
 #import "LPIsWebView.h"
 #import "LPWebViewProtocol.h"
 #import "LPCocoaLumberjack.h"
+#import "LPJSONUtils.h"
 
 @implementation LPScrollOperation
 
@@ -60,7 +61,7 @@
 
     [sv setContentOffset:point animated:YES];
 
-    return target;
+    return [LPJSONUtils jsonifyObject:target];
   } else if ([LPIsWebView isWebView:target]) {
     NSString *scrollJS = @"window.scrollBy(%@,%@);";
     if ([@"up" isEqualToString:dir]) {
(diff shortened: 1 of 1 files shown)


sapieneptus:
@jmoody I actually noticed the other day UITests involving scrolling in webviews were all failing as well. However... I think that may have been android, so I wonder if there isn't a similar issue there? I'll have to check. 

Do you know which run of CalWebApp corresponds to this PR? Can you post the link?

jmoody:
@sapieneptus [XTC Run](https://testcloud.xamarin.com/test/calwebview-cal_c08f3089-bee4-474a-80f0-d66a1820af88/) with the fix in place.

sapieneptus:
Ok, looks good generally, I'm just a bit curious about [this device](https://testcloud.xamarin.com/test/calwebview-cal_c08f3089-bee4-474a-80f0-d66a1820af88/?step=2_7_1&device=apple_ipad_air-7.1.1).

jmoody:
> this device

More details:

```
RuntimeError - Could not parse: [""] raised error: 
A JSON text must at least contain two octets! 
```

The reason that iOS 7 device is failing is because the test is bad.

* UIWebView should return `[""]`
* WKWebView should return `["{\"error\":\"A JavaScript exception occurred\",\"javascript\":\"document.javascriptMethodDoesnotExist('peach')\"}"]`

On iOS 7, there is no WKWebView so the WKWebView tests run in a UIWebView.

jmoody:
Passing: https://testcloud.xamarin.com/test/calwebview-cal_15d833cc-b7fb-4d6b-bee6-2b48d0d2606a/

PR #326, opened 2016-01-25
Add DE localization to LPTestTarget

jmoody:
### Motivation

Demonstrates that launching simulators with different locales and languages works.

**Launch simulator with locale and language** [#367](https://github.com/calabash/run_loop/pull/367)


jmoody:
Jenkins is down.

PR #332, opened 2016-02-16
Bonjour: create a unique service instance name

jmoody:
### Motivation

Having a unique instance name  makes automatic Bonjour service discovery possible.


jmoody:
Jenkins failed because of codesigning which has been fixed.

PR #134, opened 2015-03-05
Add support for OpenGL views in LPScreenshotRoute

gKamelo:
Usage of latest UIKit API to capture more complex views like OpenGL, it could be helpful for MapKit or SpriteKit  views.

PR #334, opened 2016-03-01
Screenshot: use drawViewHierarchy when possible

jmoody:
### Motivation

Resolves:

* screenshot does not capture all views; e.g. custom keyboards and OpenGL views [#789](https://github.com/calabash/calabash-ios/issues/789)
* screenshot route should use drawViewHierarchyInRect:afterScreenUpdates: for iOS 7 [#347](https://github.com/calabash/calabash-ios/issues/347)
* iOS - OpenGL views don't show up in screenshots taken locally (device or simulator), but they do in screenshots taken in the cloud [#134](https://github.com/xamarinhq/test-cloud-frameworks/issues/134)

We've known for a while that the screenshot route does not capture OpenGL - @Clancey demonstrated this at the last Xummit (back in April 2014).  We've been reluctant to change the behavior because of the quality of screenshots when UIAlertView and UIActionSheet are showing.  It seems that this problem has been resolved - at least on iOS 9 simulators and devices.

Can you review Chris?

### DIFF
--- a/calabash/Classes/FranklyServer/Routes/LPScreenshotRoute.m
+++ b/calabash/Classes/FranklyServer/Routes/LPScreenshotRoute.m
@@ -32,7 +32,14 @@
 
 - (NSData *) takeScreenshot {
 
-  CGSize imageSize = [[UIScreen mainScreen] bounds].size;
+
+  // Available on iOS >= 7
+  SEL selector = @selector(drawViewHierarchyInRect:afterScreenUpdates:);
+  UIView *view = [[UIView alloc] initWithFrame:CGRectZero];
+  BOOL drawViewHierarchy = [view respondsToSelector:selector];
+
+  CGRect bounds = [[UIScreen mainScreen] bounds];
+  CGSize imageSize = bounds.size;
   UIGraphicsBeginImageContextWithOptions(imageSize, NO, 0);
 
   CGContextRef context = UIGraphicsGetCurrentContext();
@@ -52,8 +59,12 @@
                             -[window bounds].size.width * [[window layer] anchorPoint].x,
                             -[window bounds].size.height * [[window layer] anchorPoint].y);
 
-      // Render the layer hierarchy to the current context
-      [[window layer] renderInContext:context];
+      if (drawViewHierarchy) {
+        [window drawViewHierarchyInRect:bounds afterScreenUpdates:YES];
+      } else {
+        // Render the layer hierarchy to the current context
+        [[window layer] renderInContext:context];
+      }
 
       // Restore the context
       CGContextRestoreGState(context);


calabash/Classes/FranklyServer/Routes/LPScreenshotRoute.m
```diff
@@ -32,7 +32,14 @@ - (BOOL) supportsMethod:(NSString *) method atPath:(NSString *) path {
 
 - (NSData *) takeScreenshot {
 
-  CGSize imageSize = [[UIScreen mainScreen] bounds].size;
+
+  // Available on iOS >= 7
+  SEL selector = @selector(drawViewHierarchyInRect:afterScreenUpdates:);
+  UIView *view = [[UIView alloc] initWithFrame:CGRectZero];
+  BOOL drawViewHierarchy = [view respondsToSelector:selector];
```
jmoody: I thought about asking the `[LPTouchUtils applicationWindows][0]` window if it responded to that selector, but decided this was a reasonable way to test for availability.  I am, on the other hand, suffering from vertigo and feeling super muddled.  @sapieneptus What say you?

jmoody:
@sapieneptus Please don't merge.  I am waiting for feedback from some OSS users.

- [ ] @MortenGregersen
- [ ] @kaorijp

calabash/Classes/FranklyServer/Routes/LPScreenshotRoute.m
sapieneptus: Seems aight. If a UIView responds, UIWindow should ( I mean, it might not if they're doing something goofy with invocation forwards ) but I don't see why it wouldn't. 

screen shot 2016-03-02 at 8 45 26 am

That said, if you're eventually calling `drawViewHierarchyInRect:afterScreenUpdates:` on the window, why not check on the window itself? That's all that matters, right?

jmoody: Screenshotting is slow.  I wanted avoid a extra message for each window.  I suppose there are only going to 4 or so windows max.

kaorijp:
@jmoody 
I would like report my testing result as below.
```
1. $ calabash.framework/Resources/version  
0.18.1.pre1
```  
-> Everything is working fine with this version 0.18.1.pre1.   
Screenshot is very nice !

**After using latest framework**
```
2. $ calabash.framework/Resources/version 
0.18.2.pre1
```
-> Just using this framework to make new .ipa file and run calabash.
But screenshot can not be taken away. The logs say that seems has some problem with connect to server.
Here is detail
```
Retrying.. Errno::ECONNREFUSED: (Connection refused - connect(2) (10.42.12.153:37265))
Retrying.. Errno::ECONNREFUSED: (Connection refused - connect(2) (10.42.12.153:37265))
Failing... Errno::ECONNREFUSED
    Then I take screenshot save to "login/scenario4"                          # features/step_definitions/common_steps.rb:3
      I take a screenshot
      Connection refused - connect(2) (10.42.12.153:37265) (Errno::ECONNREFUSED)
```

**I can make sure I just only changed the calabash.framework**

kaorijp:
By the way, I have a request for taking screenshot function so far. 
I'm confusing about size of screenshot ( aaa X bbb pixel) or capacity of image ( mb, kb...)
Actually , I'm using SamSung S6 for testing. So the capacity of screenshot is very large, the resolution is very high. But infact, I do not need the high quality screenshot. 

**So, can you provide a argument/parameter which allow me to specify the screen size**
It will be perfect :+1:

jmoody:
@kaorijp  On the iOS side there is no way to control the resolution.  We _might_ be able to support this, but I doubt we will want to.

I don't know if this can be supported about the Android side.  If you want this feature, please make an issue on the calabash/calabash-android repo.

kaorijp:
@jmoody 
Resolution of picture is OK now. Its just a my request ! Don't mine. 
Can you confirm the my first comment?

Issue #247, opened 2015-09-16
Remove the marked: API from WebView queries.


jmoody:
* **Cannot query UIWebView by accessibilityIdentifier or accessibilityLabel** [#735](https://github.com/calabash/calabash-ios/issues/735)

This is a breaking change and might need to wait for Calabash 2.0.

PR #337, opened 2016-03-14
Remove special :marked handling from WebView queries

jmoody:
### Motivation

* Cannot query UIWebView by accessibilityIdentifier or accessibilityLabel [#735](https://github.com/calabash/calabash-ios/issues/735)
* query ("webView marked:'XXX'") doesn't work [#85](https://github.com/calabash/calabash-ios/issues/)
* Remove the marked: API from WebView queries. #247

ATTN: @TobiasRoikjer Hurray!!!


jmoody:
@krukow 

If you agree with this change, please merge.  Otherwise we should discuss the future of this feature.

krukow:
@jmoody OK with me to kill as long as we note it in release notes.

PR #338, opened 2016-03-14
LPJSONUtils: set max and min float to avoid JSON parse errors

jmoody:
### Motivation

Users are reporting problems deserializing `CGFLOAT_MIN/MAX` with some JSON clients.

This new implementation forces a min/max float that is equivalent to `INT32_MIN/MAX`.  I thought I might try `INT16`, but if I recall correctly the bounds of MKMapView are way beyond these values.

- [x] @sapieneptus Can you review?  Please don't merge.  I am waiting to test this in the wild against an app that reproduces the problem.
- [ ] Test against a problem app.

ATTN: @GlennWilson @MatisseHack 

Possible fix for **Invalid number error while using the tree command in the REPL** [#742](https://github.com/xamarin/Xamarin.UITest/issues/742)

### DIFF
--- a/XCTest/Tests/Utils/LPJSONUtilsTest.m
+++ b/XCTest/Tests/Utils/LPJSONUtilsTest.m
@@ -776,10 +776,10 @@
   NSDictionary *dict = [LPJSONUtils dictionaryByEncodingView:view];
 
   XCTAssertEqual(((NSDictionary *)[dict objectForKey:@"frame"]).count, 4);
-  XCTAssertEqualObjects(dict[@"frame"][@"x"], @(CGFLOAT_MIN));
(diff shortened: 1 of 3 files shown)


sapieneptus:
Looks ok code-wise, how did you settle on `INT32_MAX`?

jmoody:
@sapieneptus

We want some fixed min/max value.  The obvious choice is to use a predefined min/max - like a type range.  I initially thought to use `int_16`.  However, I believe that MKMapView bounds/frame can be beyond the (-32,768, 37,767) range of a `int_16`.  The current min/max are `CGFLOAT_MIN/MAX` - these are what the C# JSON parser is choking on.  I asked @jonstoneman to run the `INT32` min/max through the C# JSON parser to verify that it can consume and produce these values.

jmoody:
No problem app is forthcoming, so I am merging.

PR #342, opened 2016-03-17
LPVersionRoute: remove reference to calabash_version

jmoody:
### Motivation

I believe this is for Frank support, but there is no reason why there needs to be a separate `version` route for Frank (it was `calabash_version`).

The goal here is to remove the CALABASH_VERSION_PATH from the client.

It is referenced twice in the client:

* https://github.com/calabash/calabash-ios/blob/develop/calabash-cucumber/bin/frank-calabash#L70
* https://github.com/calabash/calabash-ios/blob/develop/calabash-cucumber/lib/calabash-cucumber/launcher.rb#L601

The ultimate goal is to reduce the number of lines of code in client so I can implement an XCUITest action model.


krukow:
LGTM

PR #349, opened 2016-04-15
Response: set the content type to application/json

jmoody:
###  Motivation

**Some characters in an element's text/label show up as "?"** [UITest #879](https://github.com/xamarin/Xamarin.UITest/issues/879)

Also set the charset to UTF-8.

@jonstoneman I think the only way to test this in UITest is to build a new CalabashBinding.  Chris or I can show you how to generate the binding.

### TODO

- [x] Needs test

ATTN: @sapieneptus We should check the response header in DeviceAgent.

### DIFF
--- a/calabash/Classes/FranklyServer/LPCORSResponse.m
+++ b/calabash/Classes/FranklyServer/LPCORSResponse.m
@@ -1,3 +1,7 @@
+#if ! __has_feature(objc_arc)
+#warning This file must be compiled with ARC. Use -fobjc-arc flag (or convert project to ARC).
+#endif
+
 //
 //  LPCORSResponse.m
 //  LPSimpleExample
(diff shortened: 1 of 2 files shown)


jmoody:
Ready for review.

sapieneptus:
I'd like to understand why UTF-8? Don't non-ascii languages like Japanese and Chinese rely on UTF-16?  Is there an advantage to limiting the bits?

jmoody:
I think all Japanese and Chinese characters have a UTF-8 equivalent.  That is to say, every character can be represented.  In recent memory, I have interacted with one Japanese and one Chinese user around encodings - UTF-8 was good enough.  With that said, I am not an encoding expert.  I have had to deal with UTF 16 bit `.strings` files and wow, what a pain.

The problem with sending UTF-16 is that clients then need to consume UTF-16 which they probably don't by default.  Ruby, for example, does not.  There will probably come a time with UTF-16 is the right choice. I don't think it is today.  Let's try to get out of the ASCII mud. :)

sapieneptus:
I'm not really well versed on encodings, so I have to trust your judgement on that. I just would hate to use utf-8 if we knew it would cause problems for east-asian/middle eastern users, but if those character sets can be represented in utf-8 then it sounds like we should proceed with that.

calabash/Classes/FranklyServer/LPCORSResponse.m
```diff
@@ -1,3 +1,7 @@
+#if ! __has_feature(objc_arc)
+#warning This file must be compiled with ARC. Use -fobjc-arc flag (or convert project to ARC).
```
sapieneptus: I don't miss having to write that!

jmoody: Migrate project to ARC so all new files are ARC by default #266

Pull requests welcome. :)

PR #352, opened 2016-05-07
Increment LPTestTarget CFBundleVersion for every build

jmoody:
### Motivation

I want to be able to support this use case:

```
Given I testing against a physical device
And the app I am testing is already installed
When the installed app version is different than the .ipa I have on disk
Then I want Calabash to install the .ipa from disk
```

On iOS Simulators, we do a checksum to determine if an installed .app is stale.  This is not possible on physical devices; the only information about an app we can get is from the Info.plist.  The idea is to ask "Are the versions of the installed app and the app on disk the same?" and reinstall if the answer is, "No.".

### DIFF
--- a/bin/xcode-build-phase/increment-build-number.sh
+++ b/bin/xcode-build-phase/increment-build-number.sh
@@ -1,0 +1,36 @@
+#!/usr/bin/env bash
+
+set -e
+
+HEADER="${PROJECT_DIR}/${INFOPLIST_PREFIX_HEADER}"
+PLIST="${PROJECT_DIR}/${INFOPLIST_FILE}"
+touch "${PLIST}"
+
+if [ -f "${HEADER}" ]; then
+  set +e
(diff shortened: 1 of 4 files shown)


bin/xcode-build-phase/increment-build-number.sh
```diff
@@ -0,0 +1,36 @@
+#!/usr/bin/env bash
+
+set -e
+
+HEADER="${PROJECT_DIR}/${INFOPLIST_PREFIX_HEADER}"
```
sapieneptus: @jmoody Where do you get these ENV vars from? I'm not doubting them, I've just always wanted to know :)

jmoody: Two ways:

### Copy and Paste

Add an empty Run Script build phase or choose an existing Run Script build phase.

Turn on "Show environment in build log"

In the Build Logs, copy all messages and paste to a text editor. 

Search.

### Build Settings + Quick Help in Xcode

image

PR #354, opened 2016-05-13
WebQuery: rect x and y should indicate the top-left corner of the element

jmoody:
### Motivation

This aligns web query results with UIView results, the behavior of Calabash Android, and with user's expectations.

### TODO

- [x] Fix unit test

### DIFF
--- a/calabash/Classes/WebViewQuery/LPWebQuery.m
+++ b/calabash/Classes/WebViewQuery/LPWebQuery.m
@@ -67,6 +67,8 @@
 
   for (NSDictionary *d in queryResult) {
     NSMutableDictionary *dres = [NSMutableDictionary dictionaryWithDictionary:d];
+
+    // Inexplicably, rect.x and rect.y are the _center_.
     CGFloat center_x = [[dres valueForKeyPath:@"rect.x"] floatValue];
     CGFloat center_y = [[dres valueForKeyPath:@"rect.y"] floatValue];
 
@@ -82,9 +84,9 @@
 
       [dres setValue:[NSNumber numberWithFloat:finalCenter.x] forKeyPath:@"rect.center_x"];
       [dres setValue:[NSNumber numberWithFloat:finalCenter.y] forKeyPath:@"rect.center_y"];
-      [dres setValue:[NSNumber numberWithFloat:finalCenter.x] forKeyPath:@"rect.x"];
-      [dres setValue:[NSNumber numberWithFloat:finalCenter.y] forKeyPath:@"rect.y"];
-      
+      [dres setValue:[dres valueForKeyPath:@"rect.left"] forKeyPath:@"rect.x"];
+      [dres setValue:[dres valueForKeyPath:@"rect.top"] forKeyPath:@"rect.y"];
+
       /*
        When doing iframe queries, we need to store all of the information needed to
        recreate the iframe query in case subsequent tokens from the original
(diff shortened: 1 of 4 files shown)


jmoody:
@sapieneptus Jenkins is down.  I have checked UITest to see if it uses the x and y rect coordinates for gestures; it does not as far as I can tell.

jmoody:
My bad; there is a failing unit test.  We have test coverage after all. :)

jmoody:
* Unit test was failing on Xcode 6 only - forcing Travis to use Xcode 7.2
* Jenkins is in a bad state (bad update, bad simulator state, bad code signing env)

jmoody:
Jenkins test this please

jmoody:
Travis is failing because the jobs are not completing fast enough.  Everything is green before the time out.  I am willing to make a note of this and merge.  If we continue to have problems on Travis we can revisit.

jmoody:
@sapieneptus Can you review and merge?

calabash/Classes/WebViewQuery/LPWebQuery.m
```diff
@@ -67,6 +67,8 @@ + (NSArray *) arrayByEvaluatingQuery:(NSString *) query
 
   for (NSDictionary *d in queryResult) {
     NSMutableDictionary *dres = [NSMutableDictionary dictionaryWithDictionary:d];
+
+    // Inexplicably, rect.x and rect.y are the _center_.
```
sapieneptus: They're calculated to be that way in the server, _non_?

sapieneptus:
@jmoody This seems like a potentially significant change, as I imagine some tests may make use of the `rect` information. I assume this will then be in a `0.2.0` ?

jmoody:
@sapieneptus 

> some tests

You mean user tests?  I checked UITest and Calabash internals; we don't use these values.

I think this is actually a _bug_ in the current implementation.

ATTN @krukow Opinion?

calabash/Classes/WebViewQuery/LPWebQuery.m
jmoody: I think this is an unexpected behavior.  They should be named rect.center_x and rect.center_y.

jmoody: They are generated by calabash-js.

sapieneptus:
@jmoody It looks like a bug, but what I'm worried about is there may be user tests that rely on it working this way. It looks fine to merge, but I would recommend this constituting a 'major' release (0.2.0) as it is altering (technically fixing) a very core part of the server upon which users may be relying.

jmoody:
Ugh.  I am not ready to release 0.20.0.  The next release is a patch release.  Let's discuss when you arrive in Aarhus.  I don't disagree with you - it _might_ break backward compatibility.

jmoody:
@sapieneptus I discussed with @jescriba, @acroos, and @Lagoja today.

We think at 0.19.1 release is OK because it fixes a bug in our published API - the API itself did not change.

jmoody:
It will, of course, be noted in the release notes.

sapieneptus:
Remember that awful period of time where browsers actually tried to render HTML correctly and enforce the rules? Sites had already come to rely on improper HTML and broke as a result.

In the same spirit, I respectfully disagree in principle that it doesn't warrant a full 0.2.0 release. In practice, however, I have no idea if anyone actually is relying on those values. It might be totally fine. If it were known that this would cause mass breakage I'd argue more strongly, but in this case I'm content to just let it be known that I think it violates the spirit and principles of semantic versioning.

PR #361, opened 2016-08-02
Server: add set device orientation route

jmoody:
### Motivation

Xcode 8.0 beta 3 had a problem setting the device orientation on iOS Simulators.

This appears to be corrected in Xcode 8.0 beta 4.

This PR adds `setOrientation` route to the LPServer which can be used to set the UIDeviceOrientation of iOS Simulators.  It's behavior on physical devices is well understood - it appears to work, but the animation not great (on some devices).

I believe this is worth committing to give us another orientation option.  I don't think we should make a client implementation at this time.

```
# Valid orientations:
#
# up, top, upside down => PortraitUpsideDown
# bottom, down, portrait => Portrait
# left, landscape right => LandscapeRight
# right, landscape left => LandscapeLeft
#
# The orientation can also be an integer corresponding to UIDeviceOrientation enum values.
#
# Invalid orientations are treated as portrait.
#
# Unknown, FaceUp, FaceDown orientations are are treated as portrait.
http({:method => :post, :path => 'setOrientation', {:orientation => < orientation > })
```

Filed a radar and forum post.

* 27638336 REGRESSION: Cannot change the orientation of a simulator with XCUITest in Xcode 8.0 beta 3
* [Developer Forum](https://forums.developer.apple.com/message/158993)
* https://github.com/calabash/DeviceAgent.iOS/pull/132

### DIFF
--- a/XCTest/Tests/Routes/LPSetDeviceOrientationRouteTest.m
+++ b/XCTest/Tests/Routes/LPSetDeviceOrientationRouteTest.m
@@ -1,0 +1,178 @@
+#if ! __has_feature(objc_arc)
+#warning This file must be compiled with ARC. Use -fobjc-arc flag (or convert project to ARC).
+#endif
+
+#import <XCTest/XCTest.h>
+#import "LPSetDeviceOrientationRoute.h"
+#import "LPOrientationOperation.h"
+
+@interface UIDevice (LP_DEVICE_ORIENTATION_CATEGORY)
+
+-(void)setOrientation:(NSInteger)orientation animated:(BOOL)animated;
+@end
+
+@interface LPSetDeviceOrientationRoute (LPTEST)
+
+- (NSInteger)orientationWithDictionary:(NSDictionary *)arguments;
+- (NSInteger)orientationForString:(NSString *)string;
+- (BOOL)isValidUIDeviceOrientation:(NSInteger)orientation;
+
+@end
+
+@interface LPSetDeviceOrientationRouteTest : XCTestCase
+
+@property(atomic, strong) LPSetDeviceOrientationRoute *route;
+
+@end
+
+@implementation LPSetDeviceOrientationRouteTest
+
+- (void)setUp {
+  [super setUp];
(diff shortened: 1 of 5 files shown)


jmoody:
[JIRA](https://xamarin.atlassian.net/browse/TCFW-305)

XCTest/Tests/Routes/LPSetDeviceOrientationRouteTest.m
```diff
@@ -0,0 +1,178 @@
...
+  NSDictionary *args = @{};
+
+  expect([self.route orientationWithDictionary:args]).to.equal(1);
+}
+
+- (void)testOrientationWithDictionaryStringValue {
+  NSDictionary *args = @{
+                         @"orientation" : @"string"
+                         };
+  NSInteger expected = 10;
```
jescriba: Should this be: `NSInteger expected = (NSInteger)UIDeviceOrientationPortrait` ?

XCTest/Tests/Routes/LPSetDeviceOrientationRouteTest.m
```diff
@@ -0,0 +1,178 @@
...
+  OCMExpect([mock isValidUIDeviceOrientation:expected]).andReturn(YES);
+
+  NSInteger actual = [self.route orientationWithDictionary:args];
+  expect(actual).to.equal(expected);
+
+  OCMVerifyAll(mock);
+}
+
+- (void)testOrientationWithDictionaryNumberValue {
+  NSInteger expected = 10;
```
jescriba: Same comment as above

XCTest/Tests/Routes/LPSetDeviceOrientationRouteTest.m
jmoody: In both cases, I am mocking the output of a method to return _exactly_ what I want.

The goal to test all code paths.  

In particular: 

* "What happens if the orientation is an NSNumber?" and 
* "What happens if the orientation is an NSString?".

I did not want to use an valid orientation because I wanted to make it clear that values are mocked.

PR #364, opened 2016-09-12
Update for Xcode 8 GM and iPhone 7/7+

jmoody:
### Motivation

* Add device models for iPhone 7 and 7+
* Update sampleFactor for non-optimized / non-zoomed apps running on iPhone 6+ form factors
* Update the command line tools for Xcode 8.0

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,9 +1,7 @@
 source "https://rubygems.org"
 
-gem "run_loop", :github => "calabash/run_loop", :branch => "develop"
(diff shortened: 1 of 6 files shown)


jmoody:
> Travis

One explained failure in `make framework`.  Framework was built and staged, but the analysis step failed.  Subsequent tests used that framework.

The other failure was a Travis internal failure.

In both cases, all other tests passed - including XTC submit.

Safe to merge.

PR #368, opened 2016-09-22
Feature/generic scroll to mark

jescriba:


jonstoneman:
I'm using this in a UITest `app.ScrollTo()` test which searches for `marked` in `UITableViewCell`s which contain a `UIView` which contain a `UIView` which contain a `UILabel` whose text matches `marked`.  👍

jonstoneman:
I've not tested the triggering of events on UIScrollViews or when there are multiple UIScrollViews/UITableViews/UICollectionViews on screen at the same time.  Do we need to test either of those scenarios?

PR #370, opened 2016-09-29
Feature/update form factor table with latest iphone7 info

jmoody:
### Motivation

Touch coordinates are wrong because the iPhone 7 physical devices are not reporting as iPhone 6 and 6+ form factors.

### Testing

* [Before](https://testcloud.xamarin.com/test/sh-calaba-iphoneonly_1417836e-c8f0-4026-9481-fdaa492159e7/?step=0_0_1)
* [After](https://testcloud.xamarin.com/test/sh.calaba.iphoneonly_44cdcb79-d402-49e3-a4ba-e50cfbbc7c63/)

Anyone can review and merge.  This should be merged ASAP!!!

### DIFF
--- a/calabash/Classes/Utils/LPDevice.m
+++ b/calabash/Classes/Utils/LPDevice.m
@@ -314,7 +314,9 @@ NSString *const LPDeviceSimKeyIphoneSimulatorDevice_LEGACY = @"IPHONE_SIMULATOR_
 
     // iPhone 7/7+
     @"iPhone9,1" : @"iphone 6",
+    @"iPhone9,3" : @"iphone 6",
     @"iPhone9,2" : @"iphone 6+",
+    @"iPhone9,4" : @"iphone 6+",
 
     // iPad Pro 13in
     @"iPad6,7" : @"ipad pro",


jmoody:
Damn. Based off wrong branch.

jmoody:
All better.  Ready for review.

sapieneptus:
I assume the problem arose because we defaulted to some incorrect form factor? Is there no way to get the dimensions from the device itself, or am I missing the problem?

jmoody:
> problem

We answered the question "Are you like an iPhone 6" incorrectly.

The tables that I use (and Villar's uses) to create the map from model to form-factor were updated _after_ I cut the 0.20.0 release - new models were discovered after they started shipping.

> getting the dimension from the device itself

We can probably revise how we determine the form-factor by asking for the screen dimensions.  I think we did not do that at first because in the distant past the screen dimensions were not enough to determine the correct touch coordinates - we needed actual model numbers for some reason.

sapieneptus:
@jmoody I'm guessing it had to do with iPhone 3/4 which had the same dimensions but the 4 had retina display. 

I'd like to propose we re-evaluate the approach sometime now that non-retina _basically_ doesn't exist.

PR #381, opened 2017-04-24
Feature/type text operation

jmoody:
### Motivation

Publishing this to get it off my local drive.

* Users are again reporting "Timed out waiting for key event" errors
* @jescriba is investigating the possibility of clearing text while triggering delegates

### DIFF
--- a/calabash/Classes/FranklyServer/Routes/LPTypeStringRoute.m
+++ b/calabash/Classes/FranklyServer/Routes/LPTypeStringRoute.m
@@ -2,93 +2,125 @@
 #warning This file must be compiled with ARC. Use -fobjc-arc flag (or convert project to ARC).
 #endif
 
-#import "LPTypeStringOperation.h"
+#import "LPTypeStringRoute.h"
 #import "LPJSONUtils.h"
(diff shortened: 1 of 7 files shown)


calabash/Classes/FranklyServer/Routes/LPTypeStringRoute.m
```diff
@@ -0,0 +1,543 @@
...
+
+  NSString *textToType = data[@"string"];
+  if (!textToType) {
+    reason = [NSString stringWithFormat:@"Request is missing value for 'string' key"];
+    LPLogError(@"%@", reason);
+    return [self failureResponseWithReason:reason];
+  }
+
+
+  if (![target conformsToProtocol:@protocol(UITextInputTraits)]) {
```
jescriba: Why is this the case?

jmoody: UITextField, UITextView, and UISearchBar all conform to this protocol.

It does look like a bug because id<UIKeyInput> might not conform to this protocol.

PR #382, opened 2017-04-24
Route for clearing text that calls delegate methods and posts text changed notifications

jescriba:
**Motivation**
Currently clearing text doesn't trigger any of the appropriate delegate methods or NSNotifications - this PR attempts to address that.

**Update**  - I rebased to use the LPRoute approach.

Tested locally using `GET` on the `/clearText` route. Note - now this requires the keyboard to be open _before_ clearing text.

Accounts for possible scenarios:
**UITextField**
Calls `shouldChangeCharactersInRange` (when selecting all and cut) and `shouldClear` (when selecting the 'x' button) on delegate
Fires `UITextFieldTextDidChange` NSNotification

**UITextView**
Calls `shouldChangeTextIn`, `textViewDidChangeSelection`, `textViewDidChange` on delegate
Fires `UITextViewTextDidChange` NSNotification

**UISearchBar**
Calls `shouldChangeTextIn`, `textDidChange` on delegate
no notifications fired

**UIKeyInput**
Implements this protocol in which case it attempts to `deleteBackward` until `hasText` returns false

### TODO

- [x] Add tests. What's the most appropriate way of adding test?
- [x] Another point I was concerned with is the `while(hasText) { deleteBackward }` logic. Is there a way to do this with a timeout or what's the current best practice on that?
- [x] Shares some logic with PR #381 so we may want to refactor some if we merge both

### Tests

```
Scenario: LPServer GET /clearText
Given the TestApp has launched
And I am looking at the Misc tab
And I am looking at the Text Input page
When I use the LPServer to clear text
And there is no visible keyboard on the Text Input page
Then an error is raised about clear text requiring a first responder
Then I clear text with the LPServer in the UITextField
Then I clear text with the LPServer in the UITextView
Then I clear text with the LPServer in the UIKeyInput view
Given I am looking at the Pan tab
And I am looking at the Company table
Then I clear text with the LPServer in the UISearchBar
```

- [x] [Test Cloud](https://testcloud.xamarin.com/test/testapp_c0afb205-81db-401c-8864-5813a04f3c69/)

### DIFF
--- a/calabash/Classes/FranklyServer/Routes/LPClearTextRoute.m
+++ b/calabash/Classes/FranklyServer/Routes/LPClearTextRoute.m
@@ -7,6 +7,7 @@
 #import "LPTouchUtils.h"
 #import "LPCocoaLumberjack.h"
 #import "LPJSONUtils.h"
+#import "LPMachClock.h"
 
 @implementation LPClearTextRoute
 
@@ -83,8 +84,16 @@
     LPLogDebug(@"Target class %@ does not respond to setText checking UIKeyInput", target);
     if ([target conformsToProtocol:@protocol(UIKeyInput)]) {
       LPLogDebug(@"Attempting to clearText with deleteBackward");
-      while ([target hasText]) {
+      NSTimeInterval startTime = [[LPMachClock sharedClock] absoluteTime];
+      NSTimeInterval endTime = startTime + 20.0; // Timeout after 20s
+      while ([target hasText] && [[LPMachClock sharedClock] absoluteTime] < endTime) {
         [target deleteBackward];
+        CFRunLoopRunInMode(kCFRunLoopDefaultMode, 0.1, false);
+      }
+      if ([target hasText]) {
+        reason = [NSString stringWithFormat:@"Timed out clearing text from UIKeyInput"];
+        LPLogError(@"%@", reason);
+        return [self failureResponseWithReason:reason];
       }
     } else {
       reason = [NSString stringWithFormat:@"Target %@ does not respond to setText nor deleteBackward", target];
@@ -92,7 +101,7 @@
     }
   }
 
-  [self fireTextChangeMethodsWithDelegate:delegate withTarget:target];
+  [self postTextChangedNotificationAndCallDelegateMethodsOnTarget:target delegate:delegate];
 
   return [self successResponseWithResult:[LPJSONUtils jsonifyObject:target]];
 }
@@ -116,7 +125,7 @@
   return YES;
 }
 
-- (void)fireTextChangeMethodsWithDelegate:(id)delegate withTarget:(id)target {
+-(void)postTextChangedNotificationAndCallDelegateMethodsOnTarget:(id)target delegate:(id)delegate {
   if ([target isKindOfClass:[UITextField class]]) {
     [[NSNotificationCenter defaultCenter] postNotificationName:UITextFieldTextDidChangeNotification object:target];
   } else if ([target isKindOfClass:[UITextView class]]) {
--- a/calabash/Classes/FranklyServer/Routes/LPClearTextRoute.m
+++ b/calabash/Classes/FranklyServer/Routes/LPClearTextRoute.m
@@ -1,0 +1,140 @@
+#if ! __has_feature(objc_arc)
+#warning This file must be compiled with ARC. Use -fobjc-arc flag (or convert project to ARC).
+#endif
+
+#import "LPClearTextRoute.h"
+#import "UIScriptParser.h"
+#import "LPTouchUtils.h"
+#import "LPCocoaLumberjack.h"
+#import "LPJSONUtils.h"
+
+@implementation LPClearTextRoute
+
+- (BOOL) supportsMethod:(NSString *) method atPath:(NSString *) path {
+  return [method isEqualToString:@"GET"];
+}
+
+- (NSDictionary *)failureResponseWithReason:(NSString *)reason {
+    return @{
+        @"outcome" : @"FAILURE",
+        @"reason" : reason,
+        @"details" : @""
+    };
+}
+
+- (NSDictionary *)successResponseWithResult:(id)result {
+    return @{
+       @"outcome" : @"SUCCESS",
+       @"results" : @[result]
+    };
+}
+
+- (id)firstResponder {
+  NSString *query = @"* isFirstResponder:1 index:0";
+  UIScriptParser *parser = [[UIScriptParser alloc] initWithUIScript:query];
+  [parser parse];
+  NSArray *allWindows = [LPTouchUtils applicationWindows];
+  NSArray *results = [parser evalWith:allWindows];
+  if (!results || results.count == 0) {
+    return nil;
+  } else {
+    return results[0];
+  }
+}
+
+- (NSDictionary *)JSONResponseForMethod:(NSString *) method
+                                     URI:(NSString *) path
+                                    data:(NSDictionary *) data {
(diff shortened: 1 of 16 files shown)


calabash/Classes/FranklyServer/Operations/LPClearTextOperation.m
```diff
@@ -0,0 +1,29 @@
...
+- (void)fireNotifications:(id)target {
+  if ([target respondsToSelector:@selector(delegate)]) {
+    id delegate = [target performSelector:@selector(delegate)];
+
+    if ([delegate respondsToSelector:@selector(textFieldShouldClear:)]) {
+      [delegate performSelector:@selector(textFieldShouldClear:)];
+    }
+  }
+}
+
```
jmoody: We also need to call: `- textField:shouldChangeCharactersInRange:replacementString:`

```
@param string
The replacement string for the specified range. During typing, this parameter normally contains only the single new character that was typed, but it may contain more characters if the user is pasting text. When the user deletes one or more characters, the replacement string is empty.
```

This will work only for UITextField views.  A universal solution should work with:

* UITextField
* UITextView
* UISearchBar
* id<UIKeyInput>

calabash/Classes/FranklyServer/Operations/LPOperation.m
```diff
@@ -69,6 +70,8 @@ + (id) operationFromDictionary:(NSDictionary *) dictionary {
     operation = [[LPQueryAllOperation alloc] initWithOperation:dictionary];
   } else if ([opName isEqualToString:@"setText"]) {
     operation = [[LPSetTextOperation alloc] initWithOperation:dictionary];
+  } else if ([opName isEqualToString:@"clearText"]) {
+    operation = [[LPClearTextOperation alloc] initWithOperation:dictionary];
```
jmoody: I think this be a route that operates on the first responder.  It should not act on views that do not have keyboard visible keyboard attached (no quite the right way to say that).  If there is no first responder the route should respond with an error.

calabash/Classes/FranklyServer/Routes/LPClearTextRoute.m
```diff
@@ -0,0 +1,140 @@
...
+    if (currentText) {
+      existingText = (NSString *)currentText;
+    }
+  }
+
+  NSRange textRange = NSRangeFromString(existingText);
+  if ([target respondsToSelector:@selector(setText:)]) {
+    shouldClearText = [self shouldClearTextInRange:textRange WithDelegate:delegate withTarget:target];
+    if (shouldClearText) {
+      [target performSelector:@selector(setText:) withObject:@""];
```
jmoody: I *think* the correct behavior is to set the text to nil.  In iOS 7, there was change from `""` to `nil` or the other way around.

I think I would prefer using the LPInvoker instead of `performSelector:`.   If you run the static analyzer on this file, does this call generate an warning?

calabash/Classes/FranklyServer/Routes/LPClearTextRoute.m
```diff
@@ -0,0 +1,140 @@
...
+  } else if ([target isKindOfClass:[UITextView class]]) {
+    return [delegate textView:target shouldChangeTextInRange:range replacementText:@""];
+  } else if ([target isKindOfClass:[UISearchBar class]]) {
+    return [delegate searchBar:target shouldChangeTextInRange:range replacementText:@""];
+  }
+
+  return YES;
+}
+
+- (void)fireTextChangeMethodsWithDelegate:(id)delegate withTarget:(id)target {
```
jmoody: Suggested name change:

```
- (void)postTextChangedNotificationAndCallDelegateMethodsOnTarget:(id)target delegate:(id)delegate
```

calabash/Classes/FranklyServer/Routes/LPClearTextRoute.m
```diff
@@ -0,0 +1,140 @@
...
+    shouldClearText = [self shouldClearTextInRange:textRange WithDelegate:delegate withTarget:target];
+    if (shouldClearText) {
+      [target performSelector:@selector(setText:) withObject:@""];
+    }
+  } else {
+    // Check if it's UIKeyInput protocol
+    LPLogDebug(@"Target class %@ does not respond to setText checking UIKeyInput", target);
+    if ([target conformsToProtocol:@protocol(UIKeyInput)]) {
+      LPLogDebug(@"Attempting to clearText with deleteBackward");
+      while ([target hasText]) {
```
jmoody: You definitely need to pause after each `[target deleteBackward]` call.

```
# 0.1 is a suggestion - I think it makes sense.
[target deleteBackward];
CFRunLoopRunInMode(kCFRunLoopDefaultMode, 0.1, false);
```

Ah, right - UIKeyInput does not define a way to ask for text.

We have two options: time out or cap the number of characters.

A cap of 1000 characters is probably sufficient.

A timeout of 20 seconds is probably sufficient.

If you opt for timeout, please use:

* https://github.com/calabash/DeviceAgent.iOS/blob/develop/Server/Utilities/CBXMachClock.h

You can copy the file over and change the prefix to `LP`.

Follow this pattern for timing out:

```
    NSTimeInterval startTime = [[CBXMachClock sharedClock] absoluteTime];
    NSTimeInterval endTime = startTime + 20.0;

     while ([target hasText] && [[CBXMachClock sharedClock] absoluteTime] < endTime) {
         [self deleteBackwards];
         CFRunLoopRunInMode(kCFRunLoopDefaultMode, 0.1, false);
     }
 
   NSTimeInterval elapsed = [[CBXMachClock sharedClock] absoluteTime] - startTime;
   if ([target hasText) {
     // Log and fail
   } else {
    // Log and pass
  }
```

jmoody:
> Shares some logic with PR #381 so we may want to refactor some if we merge both

Assume that #381 _will not be merged_.  If it is ever merged, we can perform an extract refactor.

> Another point I was concerned with is the while(hasText) { deleteBackward } logic

Add comment about timing out or limiting to a number of characters.

> Add tests. What's the most appropriate way of adding test?

I don't think there is any way to add meaningful _unit_ tests.  The best we can do is add integration tests, either with Cucumber or UITest.

I was wrong about the LPTestTarget containing the required text input views.

Here is a depressing summary of the test applications that you might use.  No application has every view that is required.  CalSmoke is the closest.  There might be a Xamarin test app with _some_ of these.

I _think_ the best place is for these views is actually in the DeviceAgent TextApp because we will eventually want to know if DeviceAgent can handle interacting with these kinds of views.

* CalSmokeApp
  - [x] UITextField
  - [x] UITextField attached to number pad
  - [ ] UITextView
  - [x] id<UIKeyInput>
  - [ ] UISearchBar
  - [ ] WebView text field

* LPTestTarget (in calabash-ios-server)
  - [x] UITextField
  - [ ] UITextField attached to number pad
  - [ ] UITextView
  - [ ] id<UIKeyInput>
  - [ ] UISearchBar
  - [x] WebView text field

* DeviceAgent TestApp
  - [x] UITextField
  - [ ] UITextField attached to number pad
  - [x] UITextView
  - [ ] id<UIKeyInput>
  - [ ] UISearchBar
  - [ ] WebView text field

calabash/Classes/FranklyServer/Routes/LPClearTextRoute.m
jmoody: > `""` vs `nil`

Use `""`

> performSelector

This did not generate a warning.

jmoody:
Jenkins test this please.

PR #383, opened 2017-05-22
Prepare for Jenkins to stage artifacts to S3

jmoody:
### Motivation

Progress on automatically injecting the LPServer into applications.

### Concept

Jenkins jobs will generate 3 artifacts:

```
lastSuccessfulBuild/artifact/Products/s3/s3.json
lastSuccessfulBuild/artifact/Products/s3/s3.yml
# Contains only the libCalabashFAT.dylib URL
lastSuccessfulBuild/artifact/Products/s3/s3.txt
```

Clients can connect to those endpoints to find s3 URLs for various LPServer artifacts.

There are 3 "channels":  master, develop, and pull request.

```
http://calabash-ci.macminicolo.net:8080/jobCalabash%20iOS%20Server%20master
http://calabash-ci.macminicolo.net:8080/jobCalabash%20iOS%20Server%20develop
http://calabash-ci.macminicolo.net:8080/jobCalabash%20iOS%20Server%20PR
```

Assuming this pull request passes, a client could use the following to find the dylib URL:

```
$ curl http://calabash-ci.macminicolo.net:8080/jobCalabash%20iOS%20Server%20PR/lastSuccessfulBuild/artifact/Products/s3/s3.txt
```

Clients will fetch one of the s3 resource files and extract the dylib URL.

At the moment, nothing is uploaded to s3.  The URLs point to Jenkins artifacts.  This is a temporary solution until s3 keys are generated.  In the long term, we do not want to serve artifacts from Jenkins - there is not enough bandwidth on the machine and s3 has many advantages.

PR #379, opened 2017-04-19
Fixes finding subview in subviewWithMark

jescriba:
**Motivation**

I noticed failures in using `ScrollTo(Mark)` because `subviewWithMark` would return a `null` value even though a view with the matching mark was found (it just wasn't the _last_ subview so result would be set to null after having a non-null value) resulting in no scrolling being performed.

PR #376, opened 2017-01-17
Update shasum of the cert used to sign the calabash dylibs

jmoody:
### Motivation

The calabash developer certs expired on January 6th.  The `make dylibs` rule needs to be updated with the latest shasum of the calabash-developer.p12.

Related:

* iOS: update expired certs and add 2 devices [#27](https://github.com/calabash/calabash-codesign/pull/27)

PR #385, opened 2017-06-01
Release/0.20.6

jmoody:
### 0.20.6

* Prepare for Jenkins to stage artifacts to S3 #383
* GET /clearText: call delegate methods and post notifications after
  removing text from first responder #382
* Jenkins: force Xcode 7 CoreSimulatorService to load #380
* Fix: subviewWithMark only returns a match if the last subview has the mark #379
* Update shasum of the cert used to sign the calabash dylibs #376


jmoody:
Jenkins test this please

PR #390, opened 2017-07-19
Fix iPad Pro 10.5 touch coordinates for non-optimized apps

jmoody:
### Motivation

Completes:

* LPServer needs to detect apps and correct touch coordinates that are not optimized for iPad Pro 10.5 [VSTS](https://msmobilecenter.visualstudio.com/Mobile-Center/_workitems/edit/12641)

### Notes:

Apps that do not contain the correctly sized Launch Images and Icons are displayed in sample mode on iPhone 6 and 6+ form factors.  Apparently, the same is true for the iPad Pro 10.5" form factor.

When applications contain the correct image assets, the LPServer has 2 pixel accuracy on iPad Pro 10.5.  When testing non-optimized apps, the LPServer needs to scale the touch coordinates.

### DIFF
--- a/LPTestTarget/AppDelegate.m
+++ b/LPTestTarget/AppDelegate.m
@@ -1,4 +1,5 @@
 #import "AppDelegate.h"
+#import "MBFingerTipWindow.h"
 #import <CoreData/CoreData.h>
 #import "LPCoreDataStack.h"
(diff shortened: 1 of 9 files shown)


jmoody:
> Travis

The failures are false negatives.  I will disable the travis job in a subsequent commit because we are duplicating tests on Jenkins - both test Xcode 8.3.

PR #393, opened 2017-08-08
Enable tests to dispatch blocks and wait for their results on the main dispatch queue

Lievesley:
Use performSelectorOnMainThread instead of dispatch_sync because using a dispatch method prevents further blocks from being processed on the main queue until the initial block in which the httpResponseOnMainThreadForMethod was invoked has completed.
dispatch_sync prevents other blocks being processed in nested calls to NSRunLoop run methods within tests.


jmoody:
Jenkins test this please

jmoody:
@Lievesley Can you provide an example of what problem this change solves?

Lievesley:
@jmoody Here is a code example of a custom Calabash action that is currently broken as no blocks are dispatched on main queue until dispatch_sync block completes.
```
__block BOOL flag = NO;
dispatch_after(dispatch_time(DISPATCH_TIME_NOW, (int64_t)(1 * NSEC_PER_SEC)), dispatch_get_main_queue(), ^{
    flag = YES;    // Without the fix in this PR this line only gets called after the initial dispatch_sync completes
});
NSDate* loopUntil = [NSDate dateWithTimeIntervalSinceNow:10];
while ([loopUntil timeIntervalSinceNow] > 0 && !flag)
{
    [[NSRunLoop mainRunLoop] runMode:NSDefaultRunLoopMode beforeDate:loopUntil];
}
NSAssert(flag == YES, @"Flag was not set because main queue is blocked");
```

jmoody:
>  this line only gets called after the initial dispatch_sync completes

I think this is the behavior we want.  Can you describe how this is blocking a test?

Lievesley:
> I think this is the behavior we want. Can you describe how this is blocking a test?

Its not the behaviour that we want since it means that calling:
`[[NSRunLoop mainRunLoop] runMode:NSDefaultRunLoopMode beforeDate:loopUntil];`
does nothing useful. What we want to do is to process the main queue in this `runMode` method.

In our test scenarios we have backdoor methods that call asynchronous operations that dispatch events back to the main queue. We wait for these events to be processed on the main queue using:
`[[NSRunLoop mainRunLoop] runMode:NSDefaultRunLoopMode beforeDate:loopUntil];`
The asynchronous events in question must complete before the backdoor method returns.
This cannot happen if the main queue is blocked by the dispatch_sync call from the Calabash framework.

Using performSelectorOnMain thread enables the scenario I have just described while not breaking any other test case.
If you disagree please provide an example of why it is useful for:
`[[NSRunLoop mainRunLoop] runMode:NSDefaultRunLoopMode beforeDate:loopUntil];`
Not to work inside the backdoor method.

jmoody:
Thank you for the explanation.

calabash/Classes/FranklyServer/LPRouter.m
```diff
@@ -206,11 +206,21 @@ - (BOOL) supportsMethod:(NSString *) method atPath:(NSString *) path {
   if ([[NSThread currentThread] isMainThread]) {
     return [self httpResponseOnMainThreadForMethod:method URI:path];
   } else {
-    __weak typeof(self) wself = self;
-    __block NSObject<LPHTTPResponse> *result = nil;
-    dispatch_sync(dispatch_get_main_queue(), ^{
-      result = [wself httpResponseOnMainThreadForMethod:method URI:path];
-    });
+    // Use performSelectorOnMainThread instead of dispatch_sync because using a dispatch method prevents
```
jmoody: Please remove the code comments.

Lievesley: But if we remove the comment it's not going to be clear to a future maintainer of this code why we need to use performSelectorOnMainThread instead of dispatch_sync here. The danger is that the code could be reverted back to dispatch_sync as its not going to be immediately obvious to another developer why performSelectorOnMainThread on main thread is necessary here. Perhaps we can amend the comment to make it more concise? I look forward to your suggestions.

jmoody: In my opinion, the comment is not helpful.

Can you add a backdoor to the LPTestTarget app and a cucumber that demonstrates that this change allows for backdoors that call dispatch_after?

* https://github.com/calabash/calabash-ios-server/blob/develop/LPTestTarget/AppDelegate.m#L33
* https://github.com/calabash/calabash-ios-server/tree/develop/cucumber/features (add backdoor.feature with a Scenario for this example)

Lievesley: Yes, if I add a test the comment can be removed. This isn't something I have time to do this week though.

jmoody:
@Lievesley If you make me (@jmoody) a collaborator on your fork, I will add the backdoor test to the LPTestTarget app.

Lievesley:
Thanks! I just added you as a collaborator.

jmoody:
Jenkins test this please

jmoody:
Tests are failing on Jenkins because of a problem with develop.  Will resolve the issue in a subsequent commit on `origin/develop`.

PR #402, opened 2017-10-18
Support legacy apps displayed on iPhone X in letter box

jmoody:
### Motivation

This is a significant rewrite of the portion of the server that computes view hit points.

I moved touch-util-related code from LPTouchUtils to LPWebQuery when the only caller was in LPWebQuery.   This will, I think, make identifying code paths easier.

Since this a significant rewrite, it behooves us to run more tests than usual to ensure nothing is broken.  

Completes:

* LPServer can find touch coordinates on legacy apps displayed in Letter Box on iPhone X [VSTS](https://msmobilecenter.visualstudio.com/Mobile-Center/_workitems/edit/24753)

### Tests

- [x] iPhoneOnlyApp
  - [x] iPhone 6 form factor
  - [x] iPhone X form factor
  - [x] [Test Cloud](https://testcloud.xamarin.com/test/iphoneonly_d81de7c4-5b6c-4853-89e4-3ef76849d6c5/) - Force completed (1 device pending for > 2 hours)
- [x] CalSmoke
  - [x] iPhone 6 form factor
  - [x] iPhone X form factor
  - [x] [Test Cloud](https://testcloud.xamarin.com/test/calsmoke-cal_3ca4d31b-ecad-4b38-b3f8-01ee3dd39573) - failures are expected (caused by unrelated problems)
- [x] CalWebApp
  - [x] iPhone 6 form factor
  - [x] iPhone X form factor
  - [x] iPad Retina form factor
  - [x] iPad 9.7 form factor
  - [x] iPad 10.5 form factor
  - [x] iPad 12.9 form factor
  - [x] [Test Cloud](https://testcloud.xamarin.com/test/calwebview-cal_19f42754-c677-4448-87f0-d739fe7a220f) - 1 device out of 39 failed - page never loaded.

PR #403, opened 2017-10-18
Find touch coords for 768x1014 Web Apps displayed on iPad Pro form factors

jmoody:
### Motivation

When Cordova iPad apps are displayed on iPad Pro form factors, the touch coordinates are incorrect because the layout is happening in a 768x1024 rect rather than the native resolution.

Users can avoid this by updating their apps.

* http://www.cordovacircus.com/articles/ipad-pro-and-cordova

Completes:

* tap coordinates are incorrect for some iPad form factors [VSTS](https://msmobilecenter.visualstudio.com/Mobile-Center/_workitems/edit/21628)

Includes:

* Support legacy apps displayed on iPhone X in letter box #402

### Tests

See #402


jmoody:
@john7doe This change set includes the commits from #402 + 1 commit.

calabash/Classes/Utils/LPDevice.m
```diff
@@ -291,6 +292,8 @@ - (NSDictionary *) screenDimensions {
                         @"width" : @(size.width),
                         @"scale" : @(scale),
                         @"bounds" : boundsDict,
+                        @"bounds_portrait_height" : @(screenBoundsHeight),
```
john7doe: does it make sense to also log the nativeBounds?

calabash/Classes/Utils/LPDevice.m
```diff
@@ -512,6 +515,15 @@ - (BOOL) isLetterBox {
   }
 }
 
+- (BOOL) isIPhone10LetterBox {
+  if (![self isIPhone10Like]) { return NO; }
+  NSDictionary *dimensions = [self screenDimensions];
+  NSInteger screenModeHeight = [dimensions[@"height"] integerValue];
+  NSInteger screenBoundsHeight = [dimensions[@"bounds_portrait_height"] integerValue];
+
+  return screenModeHeight == 2001 && screenBoundsHeight == 667;
```
john7doe: Any case where `screenModeHeight > screenBoundsHeight` is per definition a letterbox case, right?

jmoody: Maybe?

I think it is possible that some apps will be scaled and not be in letter box.

calabash/Classes/Utils/LPDevice.m
jmoody: What are the native bounds?

john7doe: https://developer.apple.com/documentation/uikit/uiscreen/1617810-nativebounds

jmoody: Added.

PR #404, opened 2017-10-19
Test: force xcpretty to write build/reports/junit.xml

jmoody:
### Motivation

Completes:

* Jenkins: LPServer XCTest results are not generating xml files [VSTS](https://msmobilecenter.visualstudio.com/Mobile-Center/_workitems/edit/24755)

Ultimately the problem is that XCTest is not generating any test output to stdout so xcpretty cannot create a report.  I confirmed the same thing is happening in DeviceAgent.

We will have to rely on the success/failure of the `xcrun xcodebuild test` process.

Updated the Jenkins jobs accordingly.

### DIFF
--- a/bin/test/xctest.rb
+++ b/bin/test/xctest.rb
@@ -33,6 +33,9 @@ else
   warnings_as_errors="GCC_TREAT_WARNINGS_AS_ERRORS=NO"
 end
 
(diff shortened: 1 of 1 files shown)


jmoody:
Jenkins test this please

jmoody:
Jenkins test this please

PR #410, opened 2017-11-22
Static initializer can skip loading the embedded LPServer

jmoody:
### Motivation

If the app is launched with environment:

```
XTC_SKIP_LPSERVER_TOKEN => < identifier of embedded LPServer >
```

then the static initializer will not start the embedded LPServer.

Further, if the .app bundle includes a libCalabashFAT.dylib and is launched with the follow environment:

```
"DYLD_INSERT_LIBRARIES" => /path/to/installed/My.app/libCalabashFAT.dylib
```

the static initializer will start the _bundled_ LPServer.

This will allow us to "inject" an LPServer into an app that is already linked with an LPServer and eventually allow us to automatically "inject" an LPServer into an app with having to change the Xcode build settings (no more Linker flags).

Completes:

* LPServer: static initializer can control which server (embedded or bundled) is started [VSTS](https://msmobilecenter.visualstudio.com/Mobile-Center/_workitems/edit/28463)

Progress on:

* Test Cloud: Automatic Injection of Calabash server for Calabash tests in Test Cloud [VSTS](https://msmobilecenter.visualstudio.com/Mobile-Center/_workitems/edit/11596)

### Test

Automated test for Simulators

```cucumber
@dylib_inject @simulator @not_xtc
Feature: Start LPServer from embedded dylib
Skips starting the LPServer embedded in the application binary
in favor of starting the LPServer (dylib) embedded in the
application .app bundle.

  Scenario: Server embedded in the binary is launched      # features/dylib_injection.feature:11
    Given the app has launched                             # features/steps/shared.rb:29
    Then the server identifier is from the embedded binary # features/steps/dylib_injection.rb:2

  @skip_embedded_server
  Scenario: Server in the .app bundle is launched         # features/dylib_injection.feature:16
    Given the app has launched                            # features/steps/shared.rb:29
    Then the server identifier is from the embedded dylib # features/steps/dylib_injection.rb:21

2 scenarios (2 passed)
4 steps (4 passed)
0m28.419s
```


calabash/Classes/LPI.mm
```diff
@@ -0,0 +1,122 @@
...
+    } else {
+      *t = '\0';
+      i++;
+      t = ((*arr)[i]);
+    }
+    p++;
+  }
+
+  return count;
+}
```
jmoody: copy-n-paste C code.

What could go wrong?

@john7doe Can you suggest a better way of tokenizing `LPSERVERID=<SHA>` and extracting `<SHA>`?

calabash/Classes/LPI.mm
```diff
@@ -0,0 +1,122 @@
...
+
+int ___calabashserverinit();
+
+int ___lesspainfulserver = ___calabashserverinit();
+
+int ___calabashserverinit() {
```
jmoody: I discovered that depending on what Objective-C is included in this method, linking the framework will fail with:

```
"___gxx_personality_v0", referenced from: LPI.o
```

- indicating that `-lc++` linker flag is required.

calabash/Classes/LPI.mm
john7doe: Why tokenize? how about a symbol with just the sha?

jmoody: It is not enough to look for a SHA in the strings with a regex - multiple matches could be returned.

Are you suggesting something like: `LPSERVERID<SHA>` ?

john7doe: No, something like this:

#define LP_SERVER_ID_VALUE "200ea62274aea0cd35616eb9fcf3832026f7a8ea-dirty"
#define LP_SERVER_ID_KEY_VALUE "LPSERVERID="LP_SERVER_ID_VALUE

in calabash/LPGitVersionDefines.h

and then use LP_SERVER_ID_VALUE when comparing

jmoody: Interesting!

Issue #226, opened 2015-08-19
Xcode 7 beta 5: failing unit tests


jmoody:
```                                                                                                                                                     [12/15935]

  testPointerArgWithAllVariablesCreatedOffMainThread, Asynchronous wait failed: Exceeded timeout of 0.1 seconds, with unfulfilled expectations: "Test".
  /Users/moody/git/calabash/calabash-ios-server/XCTest/Tests/Invoker/LPInvokerThreadingTest.m:202


  [self waitForExpectationsWithTimeout:0.1 handler:^(NSError *error) {
    if (error) { XCTFail(@"Expectation Failed with error: %@", error); }


LPWebQueryTest
  testArrayByEvaluatingNoVisibleCenterIsZeroZeroButInsideYES, failed: caught "NSInvalidArgumentException", "-[__NSArrayM ancestorSharedWithLayer:]: unrecognized selecto$ sent to instance 0x7fd674d33e00"
  /Users/moody/git/calabash/calabash-ios-server/calabash/Classes/WebViewQuery/LPWebQuery.m:76

    CGPoint center = CGPointMake(webViewPageOffset.x + center_x, webViewPageOffset.y + center_y);
    CGPoint windowCenter = [window convertPoint:center fromView:webView];
    CGPoint keyCenter = [frontWindow convertPoint:windowCenter fromWindow:window];


  testArrayByEvaluatingNoVisibleCenterNotZeroZeroButPointInsideNO, failed: caught "NSInvalidArgumentException", "-[__NSArrayM ancestorSharedWithLayer:]: unrecognized se$ector sent to instance 0x7fd674d4f5d0"
  /Users/moody/git/calabash/calabash-ios-server/calabash/Classes/WebViewQuery/LPWebQuery.m:76
  ```
    CGPoint center = CGPointMake(webViewPageOffset.x + center_x, webViewPageOffset.y + center_y);
    CGPoint windowCenter = [window convertPoint:center fromView:webView];
    CGPoint keyCenter = [frontWindow convertPoint:windowCenter fromWindow:window];
  ```

  testArrayByEvaluatingQueryCss, failed: caught "NSInvalidArgumentException", "-[__NSArrayM ancestorSharedWithLayer:]: unrecognized selector sent to instance 0x7fd674d46
990"
  /Users/moody/git/calabash/calabash-ios-server/calabash/Classes/WebViewQuery/LPWebQuery.m:76

[...]

PR #417, opened 2018-04-09
Test: update Specta, Expecta, and OCMock

jmoody:
### Motivation

I was running some tests on Friday afternoon and discovered that Specta was breaking.

Updated Carthage to use my fork:

* ExampleGroups: do not create shared example groups in initialize [#226](https://github.com/specta/specta/pull/226)

We don't use example groups so this fix will work for us.


jmoody:
@polyakrecords 

We use Carthage to manage 3 testing frameworks:

* Specta: rspec style block-based tests for Objective-C
* Expecta: rspec style expectations
* OCMock: mock objects

jmoody:
@polyakrecords When you get a chance, please review.  There is not much here.  I just want you to see that we have these three XCTest dependencies.

PR #420, opened 2018-04-27
Server: namespace LPMachClock constant

jmoody:
### Motivation

Add the `_lp` to symbol to avoid duplicate symbol conflicts when linking the calabash.framework.


jmoody:
CI is broken broken broken because we have not updated our scripts to submit to AppCenter.

PR #431, opened 2018-09-14
Added new devices to make LPServer works with Xcode 10 GM / iOS 12

AndreyMitsyk:
**Changes:**
- Added new devices.
- Added utils.
- Added tests.
- Fixed failed tests.

Related PR: https://github.com/calabash/DeviceAgent.iOS/pull/296

### DIFF
--- a/calabash/Classes/Utils/LPDevice.h
+++ b/calabash/Classes/Utils/LPDevice.h
@@ -39,8 +39,11 @@ extern NSString *const LPDeviceSimKeyVersionInfo;
 - (BOOL) isIPhone4Like;
 - (BOOL) isIPhone5Like;
 - (BOOL) isLetterBox;
-- (BOOL) isIPhone10Like;
-- (BOOL) isIPhone10LetterBox;
+- (BOOL) isIPhoneXLike;
+- (BOOL) isIPhoneXSLike;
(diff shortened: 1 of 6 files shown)


calabash/Classes/Utils/LPDevice.h
```diff
@@ -39,8 +39,11 @@ extern NSString *const LPDeviceSimKeyVersionInfo;
 - (BOOL) isIPhone4Like;
 - (BOOL) isIPhone5Like;
 - (BOOL) isLetterBox;
-- (BOOL) isIPhone10Like;
-- (BOOL) isIPhone10LetterBox;
+- (BOOL) isIPhoneXLike;
+- (BOOL) isIPhoneXSLike;
+- (BOOL) isIPhoneXSMaxLike;
+- (BOOL) isIPhoneXRLike;
+- (BOOL) isIPhoneXLetterBox;
```
jmoody: I want to use `isIPhone10Like` instead of `isIPhoneXLike` - the X here is confusing.  The spoken name of the phone is "iPhone ten".

We don't need `isIPhone10SLike` because the 10s has the same resolution as the 10.  The 10S Max and 10R have new resolutions.

AndreyMitsyk: *Changed.*

PR #432, opened 2018-09-17
Release/0.21.7

jmoody:
### 0.21.7

* Adds new form factors for iPhone 10S, iPhone 10S Max, and iPhone 10R #431


jmoody:
Rebooted the Jenkins machine which caused a failure here.  This is safe to merge.

Issue #436, opened 2019-02-05
arm64e support


john7doe:
When linking calabash.framework with arn64e code I get the following error: 

ignoring file Vendor/calabash.framework/calabash, missing required architecture arm64e in file Vendor/calabash.framework/calabash

PR #437, opened 2019-02-06
Include arm64e slice in framework and dylibs

jmoody:
### Motivation

Resolves:

https://github.com/calabash/calabash-ios-server/issues/436


bin/make/dylibs.sh
```diff
@@ -226,24 +228,20 @@ lipo -info "${INSTALL_DIR}/libCalabashFAT.dylib"
...
     echo "${INSTALL_DIR}/libCalabashFAT.dylib does not contain bitcode for ${1}"
     exit 1
   fi
+  set -e
 }
 
 expect_bitcode arm64
+# 64e slice does not contain LLVM segment, I am not sure why not.
+# expect_bitcode arm64e
+echo "${INSTALL_DIR}/libCalabashFAT.dylib _does not_ contain bitcode for arm64e"
```
jmoody: @john7doe Any thoughts about why the `segment LLVM` is missing for the arm64e slice?

bitcode is not strictly necessary for the dylibs - I think we could safely remove the bitcode generation (will make the binaries smaller) and remove these checks.

bitcode _is necessary_ for the calabash.framework because it is statically linked - missing bitcode in a linked library will generate an error in `xcodebuild`.  The framework contains bitcode for the arm64e slice.

john7doe: don't know why its missing. Perhaps a bug or they have finally decided that they do not need a set of bit code for each slice, but has not made the necessary changes in the tooling. Time will tell :-)

jmoody: I decided to remove the bitcode generation and checks for the dylib.  After reading the tool tip for EMBED_BITCODE and reading some comments I left for myself in the bitcode check (in the dylilb build script), I think no bitcode was being generated and the check was false positive.

jmoody:
Bundler 2.0 failed the build.

See: d1c35c20e567ed8f9fe0dad904be8b2f8b765f79

PR #438, opened 2019-04-12
Update device-model mapping

AndreyMaslennikov:
We noticed one issue in the DeviceAgent - device_info[“form_factor”] returns string “iphone11,6”.
**Changes:**
- Added new devices to formFactorMap
- Fixed iPad 9.7" mapping


AndreyMaslennikov:
There is CI issue in the PR, but i believe that our changes could not affect it.
@jmoody, could you please take a look at it?

jmoody:
calabash-ci is down, probably for good.

PR #442, opened 2019-07-31
Added Azure DevOps CI/CD

OnlyForward:
I configured build pipeline for TestCloudAgent. I added matrix for Mofave-Xcode with properties (IMAGE_POOL and XCODE_Version).

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -1,5 +1,5 @@
 variables:
-  mojaveXcode: 'Mojave-Xcode-10.2.1'
+  mainJob: 'Mojave-Xcode-10.2.1'
 jobs:
 
 - job:
@@ -8,9 +8,6 @@ jobs:
       Mojave-Xcode-10.2.1:
         IMAGE_POOL: 'macOS-10.14'
         XCODE_VERSION: '10.2.1'
-      Mojave-Xcode-10.3:
-        IMAGE_POOL: 'macOS-10.14'
-        XCODE_VERSION: '10.3'
       Mojave-Xcode-11.0:
         IMAGE_POOL: 'macOS-10.14'
         XCODE_VERSION: '11'
@@ -61,7 +58,7 @@ jobs:
 
   - script: bundle exec bin/ci/jenkins/appcenter.sh
     displayName: "test appcenter"
-    condition: and(succeeded(), eq(variables['Agent.JobName'], variables['mojaveXcode']))
+    condition: and(succeeded(), eq(variables['Agent.JobName'], variables['mainJob']))
 
   - script: bundle exec bin/test/cucumber.rb
     displayName: "exec cucumber"
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -53,6 +53,9 @@ jobs:
   - script: bundle exec bin/test/xctest.rb
     displayName: "Make xctests"
 
+  - script: bundle exec bin/ci/jenkins/appcenter.sh
+    displayName: "test appcenter"
+
   - script: bundle exec bin/test/cucumber.rb
     displayName: "exec cucumber"
 
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -1,3 +1,5 @@
+variables:
+  mojaveXcode: 'Mojave-Xcode-10.2.1'
 jobs:
 
 - job:
@@ -58,6 +60,7 @@ jobs:
 
   - script: bundle exec bin/test/cucumber.rb
     displayName: "exec cucumber"
+    condition: and(succeeded(), eq(variables['Agent.JobName'], variables['mojaveXcode']))
 
   - task: PublishTestResults@2
     displayName: 'Publish Test Results'
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -1,0 +1,54 @@
+jobs:
+
+- job:
+  strategy:
+    matrix:
+      Mojave-Xcode-10.2.1:
+        IMAGE_POOL: 'macOS-10.14'
+        XCODE_VERSION: '10.2.1'
+      Mojave-Xcode-11.0:
+        IMAGE_POOL: 'macOS-10.14'
+        XCODE_VERSION: '11'
+  pool:
+    vmImage: $(IMAGE_POOL)
+  
+  steps:
+
+  - script: sudo xcode-select --switch "/Applications/Xcode_$(XCODE_VERSION).app/Contents/Developer"
+    displayName: "Select Xcode $(XCODE_VERSION)"
+
+  - script: brew install tree
+    displayName: "Install brew dependencies"
+  
+  - script: |
+      set -e
+      git clone https://$(GitHubAccessToken)@github.com/xamarinhq/calabash-codesign.git
+      calabash-codesign/apple/create-keychain.sh
+    displayName: "Download and install keychain"
+
+  - script: make framework
+    displayName: "Make framework"
+
+  - script: make dylibs
+    displayName: "Make dylibs"
+
+  - script: MAKE_FRAMEWORK=0 make ipa-cal
+    displayName: "Make ipa-call"
+
+  - script: MAKE_FRAMEWORK=0 make app-cal
+    displayName: "Make app-call"
+
+  - script: bundle install
+    displayName: "Bundle install"
+
+  - script: bundle exec bin/test/xctest.rb
+    displayName: "Make xctests"
+
+  - script: bundle exec bin/test/cucumber.rb
+    displayName: "exec cucumber"
+
+  - task: PublishTestResults@2
+    displayName: 'Publish Test Results'
+    inputs:
+      testResultsFiles: 'cucumber/reports/junit/*.xml'
+    condition: always()
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -8,9 +8,13 @@ jobs:
       Mojave-Xcode-10.2.1:
         IMAGE_POOL: 'macOS-10.14'
         XCODE_VERSION: '10.2.1'
+      Mojave-Xcode-10.3:
+        IMAGE_POOL: 'macOS-10.14'
+        XCODE_VERSION: '10.3'
       Mojave-Xcode-11.0:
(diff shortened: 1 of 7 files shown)


azure-pipelines.yml
```diff
@@ -0,0 +1,67 @@
...
+    displayName: "Select Xcode $(XCODE_VERSION)"
+
+  - script: brew install tree
+    displayName: "Install brew dependencies"
+
+  - script: xcodebuild -version
+    displayName: "Version of xcodebuild"
+
+  - script: git clone -b develop https://github.com/calabash/FBSimulatorControl.git ../FBSimulatorControl
+    displayName: "Download FBSimulatorControl"
```
jmoody: This is unnecessary.  Remove it.

azure-pipelines.yml
```diff
@@ -0,0 +1,49 @@
...
+  - script: make framework
+    displayName: "Make framework"
+
+  - script: make dylibs
+    displayName: "Make dylibs"
+
+  - script: MAKE_FRAMEWORK=0 make ipa-cal
+    displayName: "Make ipa-call"
+
+  - script: gtimeout --foreground --signal SIGKILL 74m bundle exec bin/test/xctest.rb
```
jmoody:
```suggestion
  - script: bundle exec bin/test/xctest.rb
```

The `gtimeout --foreground` calls were necessary on Jenkins because the tests were occasionally blocking  indefinitely.  They should not be necessary here.

azure-pipelines.yml
```diff
@@ -0,0 +1,54 @@
...
+      Mojave-Xcode-11.0:
+        IMAGE_POOL: 'macOS-10.14'
+        XCODE_VERSION: '11'
+  pool:
+    vmImage: $(IMAGE_POOL)
+  
+  steps:
+
+  - script: sudo xcode-select --switch "/Applications/Xcode_$(XCODE_VERSION).app/Contents/Developer"
+    displayName: "Select Xcode $(XCODE_VERSION)"
```
jmoody: Let's add the same version and path information as we have in the iOSDeviceManager and DeviceAgent pipelines.

```      set -e
	      sudo xcode-select --switch "/Applications/Xcode_$(XCODE_VERSION).app/Contents/Developer"
	      echo "Xcode version: $(xcrun xcodebuild -version)"
	      echo "Xcode path: $(xcrun -f xcodebuild)"
	      echo "Xcode select path: $(xcode-select --print-path)"
```

bin/test/cucumber.rb
```diff
@@ -44,21 +44,26 @@
...
     else
       devices = {
         :iphone7 => 'iPhone 7',
-        :iphone7plus => 'iPhone 7 Plus'
+        :iphone7Plus => 'iPhone 7 Plus'
       }
     end
 
-    RunLoop::CoreSimulator.quit_simulator
+    RunLoop::CoreSimulator.terminate_core_simulator_processes
```
jmoody: Replace this with:

```
Luffa.unix_command("bundle exec run-loop simctl manage-processes")
```

bin/test/cucumber.rb
```diff
@@ -44,21 +44,26 @@
 
     sim_version = RunLoop::Version.new("#{sim_major}.#{sim_minor}")
 
-    if ENV["JENKINS_HOME"]
+    if RunLoop::Environment.jenkins?
```
jmoody: This section is still incorrec.t

Please remove the Jenkins branching.  Delete the `if/else`.

bin/test/cucumber.rb
```diff
@@ -44,21 +44,26 @@
...
       devices = {
         :iphone7Plus => 'iPhone 7 Plus',
         :air => 'iPad Air 2',
         :iphoneSE => 'iPhone SE',
         :iphone7 => 'iPhone 7'
       }
+    elsif RunLoop::Environment.azurepipelines?
+      devices = {
```
jmoody: When running in the pipeline, `devices` should contain these:

```
iPhone SE
iPhone Xs
iPhone Xr
iPhone Xs Max
iPad Pro 10.5-inch
iPad Pro 9.7-inch
iPhone 8 Plus
iPhone 8
iPhone X
```

bin/test/cucumber.rb
```diff
@@ -44,21 +44,26 @@
...
+        :iphoneXs => 'iPhone Xs',
+        :iphoneXsMax => 'iPhone Xs Max'
+      }
     else
       devices = {
         :iphone7 => 'iPhone 7',
-        :iphone7plus => 'iPhone 7 Plus'
+        :iphone7Plus => 'iPhone 7 Plus'
       }
```
jmoody: Locally, `devices` should contain an iPhone Xs and iPhone Xr.

azure-pipelines.yml
```diff
@@ -0,0 +1,63 @@
...
+      echo "Xcode version: $(xcrun xcodebuild -version)"
+      echo "Xcode path: $(xcrun -f xcodebuild)"
+      echo "Xcode select path: $(xcode-select --print-path)"
+    displayName: "Select Xcode $(XCODE_VERSION)"
+
+  - script: |
+      set -e
+      bundle install
+      ruby -v
+      echo "install: --no-document --env-shebang\nupdate:  --no-document --env-shebang" > ~/.gemrc
```
jmoody: This needs to happen _before_ the `bundle install`.

bin/test/cucumber.rb
```diff
@@ -51,10 +51,22 @@
...
+      if xcode_version.major < 11
+        devices = {
+          :iPhoneXs => 'iPhone Xs',
+          :iphoneXr => 'iPhone Xʀ',
+        }
+      else
+        devices = {
+          :iPhoneXs => 'iPhone Xs',
+          :iphoneXr => 'iPhone Xʀ',
+        }
```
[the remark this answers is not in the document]
OnlyForward: Could you please confirm that we should add all these devices ?

bin/test/cucumber.rb
max-lobanov: @jmoody , could you please confirm that we should add all these devices for testing

bin/test/cucumber.rb
```diff
@@ -51,10 +51,22 @@
         :iphoneSE => 'iPhone SE',
         :iphone7 => 'iPhone 7'
       }
+    elsif ENV["AGENT_VERSION"]
+      if xcode_version.major < 11
```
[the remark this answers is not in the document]
max-lobanov: It was outdate changes. For now it is removed

cucumber/features/steps/dylib_injection.rb
```diff
@@ -70,7 +70,7 @@ def installed_test_target_app_cal_dylib
 When(/^running in App Center the entitlement injector is loaded$/) do
   if RunLoop::Environment.xtc?
     mark = "Tomato: promoted to vegetable"
-    timeout = 10
+    timeout = 120
```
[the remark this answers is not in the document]
max-lobanov: It is outdated changes. For now it is reverted.

bin/test/cucumber.rb
jmoody: Yes.  Please add those devices.

jmoody:
@OnlyForward There are failing tests on Xcode 11.  Do these tests fail locally as well?

OnlyForward:
> @OnlyForward There are failing tests on Xcode 11. Do these tests fail locally as well?

No, they don't

jmoody:
In the `bin/ci/jenkins/appcenter.sh` script, the device set needs to be changed to:

```
App-Center-Test-Cloud/calabash-ios-server
```

The original device set has been deleted.

@OnlyForward @max-lobanov

azure-pipelines.yml
```diff
@@ -0,0 +1,66 @@
...
+  - script: MAKE_FRAMEWORK=0 make ipa-cal
+    displayName: "Make ipa-call"
+
+  - script: MAKE_FRAMEWORK=0 make app-cal
+    displayName: "Make app-call"
+
+  - script: bundle exec bin/test/xctest.rb
+    displayName: "Make xctests"
+
+  - script: bundle exec bin/ci/jenkins/appcenter.sh
```
jmoody: We need to change this step to only submit a job to Test Cloud for 1 image pool / Xcode combination.

When the image pool is `macOS-10.14` and the Xcode is `10.2.1`, submit a test.  Otherwise, write a log message that the step is being skipped.

bin/test/cucumber.rb
```diff
@@ -110,7 +141,7 @@
     exit 0 if failed == 0
 
     # the travis ci environment is not stable enough to have all tests passing
-    exit failed unless Luffa::Environment.travis_ci?
+    exit failed unless RunLoop::Environment.travis? || RunLoop::Environment.azurepipelines?
```
jmoody: Remove the `RunLoop::Environment.travis?` code.  Remove also the comment.

bin/test/cucumber.rb
```diff
@@ -110,7 +141,7 @@
...
 
     # we'll take 75% passing as good indicator of health
     expected = 75
```
jmoody: Do we know what the failure rate is on ADO?

cucumber/features/steps/shared.rb
```diff
@@ -8,6 +8,22 @@ def get_http_route(path)
       body = http({:method => "GET", :path => path}, {})
       response_body_to_hash(body)
     end
+
+    def switch_to_second_tab
+      xcode = RunLoop::Xcode.new
+      if xcode.version_gte_110?
+        # Separate case for Xcode 11 beta since it doesn't switch to the second tab sometimes
+        # retry_count = 15 minutes to make sure that simulator has stabilized
+        retry_count = 900
```
jmoody: Wow.  That is a long time to retry.  Don't change.

Here is good place where this test should just fail when the Xcode 11 version is out of beta.

Not blocking. Nothing to do.

bin/test/cucumber.rb
max-lobanov: - Xcode 10.2.1: passrate is 100%
- Xcode 11: we run 6 devices and most often all of them pass successfully.
But sometimes 1 of devices failed so failure rate 1/6 = 16%. I think 75-80% should be ok

PR #446, opened 2019-08-06
Fix logic to switch to the second tab

max-lobanov:
# Issue
2 scenarios fail on CI randomly:
1) `Scenario: Entitlement Injector has been loaded`
2) `Scenario: Touch small buttons`
Both of them fail about 4-5 times per 10 runs

# Debug process
The main issues with both tests is that they can't get some value by query
I have added request `query(*)` to failing tests to determine what elements are showed on device screen. I have found that test app still shows elements from the first tab after step `I go to the second tab`. So this step doesn't switch tabs sometimes.
The main issue is that test app sometimes is not switched to the second tab randomly.

# Fix
I have added logic to touch second tab icon until the second tab is shown.
```
while query("view marked: 'Second View'").empty?
    touch("UITabBarButton index:1")
  end
```

# How it was tested
I have run 11 builds in CI and all of them passed without issue. Both tests were green.

I am not sure that such retry logic is good solution but it works and proofs main idea.
The root cause can be related to some instability and hangs of Xcode 11 simulators since Xcode 11 is still `beta`
@jmoody , probably you can advice better solution.

### DIFF
--- a/cucumber/features/steps/shared.rb
+++ b/cucumber/features/steps/shared.rb
@@ -9,10 +9,20 @@ module LPTestTarget
       response_body_to_hash(body)
     end
 
-    def switch_to_second_tab_retryable
-      begin
+    def switch_to_second_tab
+      xcode = RunLoop::Xcode.new
+      if xcode.version_gte_110?
+        # Separate case for Xcode 11 beta since it doesn't switch to the second tab sometimes
+        # retry_count = 15 minutes to make sure that simulator has stabilized
+        retry_count = 900
+        begin
+          touch("UITabBarButton index:1")
+          retry_count = retry_count - 1
+          sleep(1)
+        end while retry_count > 0 && query("view marked: 'Second View'").empty?
+      else
         touch("UITabBarButton index:1")
-      end while query("view marked: 'Second View'").empty?
+      end
     end
   end
 end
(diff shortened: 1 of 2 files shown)


cucumber/features/steps/shared.rb
```diff
@@ -8,6 +8,12 @@ def get_http_route(path)
       body = http({:method => "GET", :path => path}, {})
       response_body_to_hash(body)
     end
+
+    def switch_to_second_tab_retryable
+      begin
+        touch("UITabBarButton index:1")
+      end while query("view marked: 'Second View'").empty?
```
jmoody: I like the solution.

Let's do two things:

1.  avoid an infinite loop by putting in a counter.  Let's try 10 times.  And have a sleep of 1 second between each try.

2. Add documentation about why we are doing this.
3. Add a branch - only do this when Xcode >= 11.  I think you can also ask run loop if the Xcode is beta.  So If Xcode >= 11 && Xcode is beta.

max-lobanov: 1) Done, but I have to set bigger timeout to make sure that build is absolutely stable.
2) Done
3) Done, I have added condition for Xcode 11 but `beta` condition doesn't work in CI since we remove `-beta` occurrence from xcode name

max-lobanov:
I have resolved comments and going to merge the PR. Let's review all changes in the original PR: https://github.com/calabash/calabash-ios-server/pull/442

cucumber/features/steps/shared.rb
jmoody: Why do you remove `-beta` from the Xcode name?  That seems like a bad choice.

PR #447, opened 2019-08-07
Publish TestCloudAgent artifacts to blob storage

ivan-nosar:
This PR adds new build step to the `azure-pipelines.yml` file. This build step will perform publishing of build artifacts (see exact list of artifacts below) to the [Azure Blob storage](https://ms.portal.azure.com/#%40microsoft.onmicrosoft.com/resource/subscriptions/9e13100f-ca82-4813-b6ea-9c8d72655924/resourceGroups/TestCloud/providers/Microsoft.Storage/storageAccounts/xtcruntimeartifacts/containersList) (Container name: `ios-test-cloud-agent`). Artifacts from build job with macOS Mojave and Xcode 10.2.1 would be published.

List of artifacts:
```
calabash.framework.zip
calabash-dylibs/libCalabashFAT.dylib
calabash-dylibs/libCalabashARM.dylib
calabash-dylibs/libCalabashSim.dylib
calabash-dylibs/Headers.zip
```

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -1,6 +1,6 @@
 variables:
   # `mainJob` variable points to the matrix case that will be the main case
-  # it will run `appcenter-cli` test step and publish artifacts to the Azure Blob
+  # it will run `appcenter-cli` test step
   mainJob: 'Mojave-Xcode-10.2.1'
 jobs:
 
@@ -73,7 +73,7 @@ jobs:
       AZURE_STORAGE_CONNECTION_STRING: $(azureStorageConnectionString)
       SOURCE_BRANCH: $(Build.SourceBranch)
     displayName: "Publish to Azure Blob Storage"
-    condition: and(succeeded(), eq(variables['Agent.JobName'], variables['mainJob']))
+    condition: and(succeeded(), eq(variables['XCODE_VERSION'], variables['10.2.1']))
 
   - task: PublishTestResults@2
     displayName: 'Publish Test Results'
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -73,7 +73,7 @@ jobs:
       AZURE_STORAGE_CONNECTION_STRING: $(azureStorageConnectionString)
       SOURCE_BRANCH: $(Build.SourceBranch)
     displayName: "Publish to Azure Blob Storage"
-    condition: and(succeeded(), eq(variables['XCODE_VERSION'], variables['10.2.1']))
+    condition: and(succeeded(), eq(variables['XCODE_VERSION'], '10.2.1'))
 
   - task: PublishTestResults@2
     displayName: 'Publish Test Results'
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -1,6 +1,6 @@
 variables:
   # `mainJob` variable points to the matrix case that will be the main case
-  # it will run `appcenter-cli` test step
+  # it will run `appcenter-cli` test step and publish artifacts to the Azure Blob
   mainJob: 'Mojave-Xcode-10.2.1'
 jobs:
 
@@ -73,7 +73,7 @@ jobs:
       AZURE_STORAGE_CONNECTION_STRING: $(azureStorageConnectionString)
       SOURCE_BRANCH: $(Build.SourceBranch)
     displayName: "Publish to Azure Blob Storage"
-    condition: and(succeeded(), eq(variables['XCODE_VERSION'], '10.2.1'), startsWith(variables['Build.SourceBranch'], 'refs/tags/*'))
+    condition: and(succeeded(), eq(variables['Agent.JobName'], variables['mainJob']), startsWith(variables['Build.SourceBranch'], 'refs/tags/*'))
 
   - task: PublishTestResults@2
     displayName: 'Publish Test Results'
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -2,6 +2,8 @@ variables:
   # `mainJob` variable points to the matrix case that will be the main case
   # it will run `appcenter-cli` test step and publish artifacts to the Azure Blob
   mainJob: 'Mojave-Xcode-10.2.1'
+  # `PublishFrom` variable points to the matrix case that will publish build artifacts to the blob storage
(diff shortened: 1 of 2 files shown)


azure-pipelines.yml
```diff
@@ -65,6 +65,16 @@ jobs:
...
 
+  # azureStorage* env vars are defined in the pipeline UI as secret variables
+  - bash: "./bin/ci/az-publish.sh"
+    env:
+      AZURE_STORAGE_ACCOUNT: $(azureStorageAccount)
+      AZURE_STORAGE_KEY: $(azureStorageKey)
+      AZURE_STORAGE_CONNECTION_STRING: $(azureStorageConnectionString)
+      SOURCE_BRANCH: $(Build.SourceBranch)
+    displayName: "Publish to Azure Blob Storage"
+    condition: and(succeeded(), eq(variables['Agent.JobName'], variables['mainJob']))
```
max-lobanov: We need additional condition for the branch. I think artifact publishing should be performed only for `master` / `develop` branch.
We can use the current condition for testing but please add branch condition before merging

max-lobanov: @jmoody

bin/ci/az-publish.sh
```diff
@@ -0,0 +1,75 @@
...
+set -eo pipefail
+CONFIG=Debug
+
+# $1 => SOURCE PATH
+# $2 => TARGET NAME
+function azupload {
+    echo "Dry run: ${1} published with name: ${2}"
+# az storage blob upload \
+#   --container-name ios-test-cloud-agent \
+#   --if-none-match "*" \
```
jmoody: I think it is acceptable to overwrite if there is an existing file.

bin/ci/az-publish.sh
```diff
@@ -0,0 +1,75 @@
+#!/usr/bin/env bash
+
+source bin/ditto.sh
+
+set -eo pipefail
+CONFIG=Debug
```
jmoody: What does this do?

bin/ci/az-publish.sh
```diff
@@ -0,0 +1,75 @@
...
+# Upload `calabash.framework.zip`
+CALABASH_FRAMEWORK="${WORKING_DIR}/calabash.framework.zip"
+zip_with_ditto "${WORKING_DIR}/calabash.framework" "${CALABASH_FRAMEWORK}"
+CALABASH_FRAMEWORK_NAME="calabash.framework-${VERSION}-${GIT_SHA}.zip"
+azupload "${CALABASH_FRAMEWORK}" "${CALABASH_FRAMEWORK_NAME}"
+echo "${CALABASH_FRAMEWORK} artifact uploaded with name ${CALABASH_FRAMEWORK_NAME}"
+
+# Upload `libCalabashFAT.dylib`
+CALABASH_FAT="${WORKING_DIR}/calabash-dylibs/libCalabashFAT.dylib"
+CALABASH_FAT_NAME="libCalabashFAT-${VERSION}-${GIT_SHA}.dylib"
```
jmoody: I think we need to have another piece of information:  the Xcode version that compiled the artifact.

The bin/xcode.sh script has a function we can use `xcode_version`

```
source bin/xcode.sh

...
XC_VERSION=$(xcode_version)
```

azure-pipelines.yml
jmoody: Can we change this to explicitly branch on the Xcode version?

Like this maybe?

```
eq(variables['XCODE_VERSION'], '10.2.1')
```

bin/ci/az-publish.sh
```diff
@@ -0,0 +1,75 @@
...
+az --version
+
+WORKING_DIR="${BUILD_SOURCESDIRECTORY}"
+
+# Upload `calabash.framework.zip`
+CALABASH_FRAMEWORK="${WORKING_DIR}/calabash.framework.zip"
+zip_with_ditto "${WORKING_DIR}/calabash.framework" "${CALABASH_FRAMEWORK}"
+CALABASH_FRAMEWORK_NAME="calabash.framework-${VERSION}-${GIT_SHA}.zip"
+azupload "${CALABASH_FRAMEWORK}" "${CALABASH_FRAMEWORK_NAME}"
+echo "${CALABASH_FRAMEWORK} artifact uploaded with name ${CALABASH_FRAMEWORK_NAME}"
```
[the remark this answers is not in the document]
ivan-nosar: Done

bin/ci/az-publish.sh
ivan-nosar: Done

bin/ci/az-publish.sh
ivan-nosar: This line was copied from `Xamarin.UITest/nunit-extensions/nunit3/az-publish.sh`. Forgot to remove right away. Removed.

bin/ci/az-publish.sh
ivan-nosar: Done

azure-pipelines.yml
ivan-nosar: Sure, done.
@jmoody what do you think, should we explicitly add the macOS version check in condition too?

jmoody: Can we configure this step to only publish when the branch is a tag?

https://github.com/calabash/calabash-ios-server/tags

e.g. `tag/0.21.8` ?

@max-lobanov

jmoody: The condition should use the Xcode version explicitly.

max-lobanov: Honestly, I prefer `mainJob` approach because we will need to change only 1 place when the default configuration is updated (for example, when Xcode 10.3 is released)

With this approach we will have clear separation: `main job` and `additional jobs`.
Main job runs all tasks including publishing to the Azure Blob, run `appcenter-cli` and etc.
Additional jobs run only testing.

jmoody: I agree.  I missed the addition of the "main job" variable.  This is a good idea.

I was worried the "main job" was an azure pipeline concept meaning "the first job in the matrix" or something.

ivan-nosar: Returned condition with `mainJob`

ivan-nosar: Added condition for tags:
```
eq(variables['Agent.JobName'], variables['mainJob'])
```

PR #450, opened 2019-08-15
Changes to use Key Vault in yaml file

owenniblock:


jmoody:
Tests did pass on manual rerun after authorizing the Variable Group connection in the UI

PR #453, opened 2019-10-09
Add develop artifacts

AndreyMaslennikov:
In this PR we have extended `bin/ci/az-publish.sh` to create and upload `{develop|release}.txt` and `{develop|release}.json` files to the Azure blob container.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -90,7 +90,7 @@ jobs:
       AZURE_STORAGE_CONNECTION_STRING: $(azureStorageConnectionString)
       SOURCE_BRANCH: $(Build.SourceBranch)
     displayName: "Publish to Azure Blob Storage"
-    condition: and(succeeded(), eq(variables['Agent.JobName'], variables['mainJob']), startsWith(variables['Build.SourceBranch'], 'refs/tags/'))
+    condition: and(succeeded(), eq(variables['Agent.JobName'], variables['mainJob']), or(startsWith(variables['Build.SourceBranch'], 'refs/tags/'), startsWith(variables['Build.SourceBranch'], 'develop'))
 
   - task: PublishTestResults@2
     displayName: 'Publish Test Results'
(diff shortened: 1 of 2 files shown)


bin/ci/az-publish.sh
```diff
@@ -41,9 +43,13 @@ VERSION=$(xcrun strings calabash-dylibs/libCalabashFAT.dylib | grep -E "CALABASH
...
 
-az --version
+# Evaluate calabash.framework SHASUM256
+FRAMEWORK_SHASUM256=$(shasum --algorithm 256 ${WORKING_DIR}/calabash.framework/calabash | cut -d " " -f 1)
 
-WORKING_DIR="${BUILD_SOURCESDIRECTORY}"
+# Evaluate dylibFAT SHASUM256
+DYLIBFAT_SHASUM256=$(shasum --algorithm 256 ${WORKING_DIR}/calabash-dylibs/libCalabashFAT.dylib | cut -d " " -f 1)
+
+az --version
```
maxim-lobanov: Do we really need it? :)

AndreyMaslennikov: It's interesting for me as well :)
This command has been here before, and no one knows why. Should I remove it?

azure-pipelines.yml
```diff
@@ -90,7 +90,7 @@ jobs:
       AZURE_STORAGE_CONNECTION_STRING: $(azureStorageConnectionString)
       SOURCE_BRANCH: $(Build.SourceBranch)
     displayName: "Publish to Azure Blob Storage"
-    condition: and(succeeded(), eq(variables['Agent.JobName'], variables['mainJob']), startsWith(variables['Build.SourceBranch'], 'refs/tags/'))
+    condition: and(succeeded(), eq(variables['Agent.JobName'], variables['mainJob']), or(startsWith(variables['Build.SourceBranch'], 'refs/tags/'), startsWith(variables['Build.SourceBranch'], 'develop')))
```
jmoody: Let's change `mainJob` to `AzurePublishWhen:`

bin/ci/az-publish.sh
jmoody: Ha!  It is left over from when we were exploring publishing to blob storage using the command line tools.

We can remove it or leave it.

azure-pipelines.yml
AndreyMaslennikov: Done

PR #458, opened 2020-05-21
Adapt calabash-ios-server for Xcode 11.5 GM

AndreyMaslennikov:
We got the following alerts during the `calabash-ios-server` testing:

- `Enumeration values 'UITouchPhaseRegionEntered', 'UITouchPhaseRegionMoved', and 'UITouchPhaseRegionExited' not handled in switch`
We tried to add this values in the switch, but it caused failures for old Xcode versions. So we add the `default` case.

- `'UILaunchImages' has been deprecated, use launch storyboards instead`
We replaced .xib launch screen with storyboard one.


leffsu:
@AndreyMaslennikov, I propose moving these enums into the bottom so that we wouldn't trigger unexpected behaviour when phase is set to these values.

PR #460, opened 2020-08-22
Ignore text caret animation

JoeSSS:
In iOS 14 the text caret animation started to animate inside of the thread which is not expected behaviour for none animation check. This PR skips animation keys with `UITextSelectionViewCaretBlinkAnimation`

### DIFF
--- a/calabash/Classes/FranklyServer/Routes/LPConditionRoute.m
+++ b/calabash/Classes/FranklyServer/Routes/LPConditionRoute.m
@@ -205,9 +205,10 @@
                                                       BOOL *stop) {
             CAAnimation *animation = [view.layer animationForKey:key];
             if (animation.duration > kLPConditionRouteAnimationDurationLimit &&
-               ! [key  isEqual: @"UITextSelectionViewCaretBlinkAnimation"]) { 
+                ! [key  isEqual: @"UITextSelectionViewCaretBlinkAnimation"]) { 
               atLeastOneAnimating = YES;
               *stop = YES;
+            }
           }];
         }
       }
--- a/calabash/Classes/FranklyServer/Routes/LPConditionRoute.m
+++ b/calabash/Classes/FranklyServer/Routes/LPConditionRoute.m
@@ -182,7 +182,7 @@
         UIView *view = (UIView *)match;
         NSArray *animationKeys = [[view.layer animationKeys] copy];
         for (NSString *key in animationKeys) {
-          if ([key  isEqual: @"UITextSelectionViewCaretBlinkAnimation"]) {
+          if ([key isEqual: @"UITextSelectionViewCaretBlinkAnimation"]) {
             return false;
           } else {
             CAAnimation *animation = [view.layer animationForKey:key];
--- a/calabash/Classes/FranklyServer/Routes/LPConditionRoute.m
+++ b/calabash/Classes/FranklyServer/Routes/LPConditionRoute.m
@@ -204,7 +204,8 @@
                                                       NSUInteger idx,
                                                       BOOL *stop) {
             CAAnimation *animation = [view.layer animationForKey:key];
-            if ([key  isEqual: @"UITextSelectionViewCaretBlinkAnimation"]) {
+            if (animation.duration > kLPConditionRouteAnimationDurationLimit &&
+               ! [key  isEqual: @"UITextSelectionViewCaretBlinkAnimation"]) { 
               // do nothing
             } else if (animation.duration > kLPConditionRouteAnimationDurationLimit) {
               atLeastOneAnimating = YES;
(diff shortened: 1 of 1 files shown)


calabash/Classes/FranklyServer/Routes/LPConditionRoute.m
```diff
@@ -173,16 +173,21 @@ - (BOOL) atLeastOneAnimatingOnMainThreadWithQuery:(id) query {
...
   if ([[NSThread currentThread] isMainThread]) {
     NSArray *matches = [LPOperation performQuery:query];
     for (id match in matches) {
       if ([match isKindOfClass:[UIView class]]) {
         UIView *view = (UIView *)match;
         NSArray *animationKeys = [[view.layer animationKeys] copy];
         for (NSString *key in animationKeys) {
-          CAAnimation *animation = [view.layer animationForKey:key];
-          return animation.duration > kLPConditionRouteAnimationDurationLimit;
+          if ([key  isEqual: @"UITextSelectionViewCaretBlinkAnimation"]) {
```
jamuc, line 185: Minor: I think you have two spaces here, should only be one.
```suggestion
          if ([key isEqual: @"UITextSelectionViewCaretBlinkAnimation"]) {
```

calabash/Classes/FranklyServer/Routes/LPConditionRoute.m
```diff
@@ -199,7 +204,9 @@ - (BOOL) atLeastOneAnimatingOnMainThreadWithQuery:(id) query {
                                                       NSUInteger idx,
                                                       BOOL *stop) {
             CAAnimation *animation = [view.layer animationForKey:key];
-            if (animation.duration > kLPConditionRouteAnimationDurationLimit) {
+            if ([key  isEqual: @"UITextSelectionViewCaretBlinkAnimation"]) {
+              // do nothing
```
jamuc, line 208: It feels weird having an empty if-case. An alternative approach could be to extract the boolean condition into it's own method and just have an if. So you would have

```
- (BOOL) keyIsNotBlinkingAndAnimationDurationIsHigherThanConditionRouteAnimationDurationLimit {
    return ![key isEqual: @"..."] && animation.duration > klPCondi....
}
```
Is this better? I don't know. Having ridiculously long method names is the Apple way.

calabash/Classes/FranklyServer/Routes/LPConditionRoute.m
```diff
@@ -199,7 +204,9 @@ - (BOOL) atLeastOneAnimatingOnMainThreadWithQuery:(id) query {
                                                       NSUInteger idx,
                                                       BOOL *stop) {
             CAAnimation *animation = [view.layer animationForKey:key];
-            if (animation.duration > kLPConditionRouteAnimationDurationLimit) {
+            if ([key  isEqual: @"UITextSelectionViewCaretBlinkAnimation"]) {
```
jamuc, line 207: Same thing here: Extra space between `key` and `isEqual`
```suggestion
            if ([key isEqual: @"UITextSelectionViewCaretBlinkAnimation"]) {
```

jamuc (approved):
Yeah, I'm also no objective C guru. I left some comments, I don't know if they are helpful. I was just focused on code style, I have no idea what this PR changes :)

calabash/Classes/FranklyServer/Routes/LPConditionRoute.m
huNt-FMJ: if ([key  isEqual: @"UITextSelectionViewCaretBlinkAnimation"]) {

Is that more of a thing regarding documentation?

Looks like we have three conditions:
1.  if ([key  isEqual: @"UITextSelectionViewCaretBlinkAnimation"]) {
2. else if (animation.duration > kLPConditionRouteAnimationDurationLimit) && not 1.
3. (basically) else

I like @jamuc's idea, but in this case it might be fine to combine both statements without having a dedicated method:
```suggestion
            if (animation.duration > kLPConditionRouteAnimationDurationLimit &&
               ! [key  isEqual: @"UITextSelectionViewCaretBlinkAnimation"]) { 
```
since the code itself seems legible enough.

calabash/Classes/FranklyServer/Routes/LPConditionRoute.m
```diff
@@ -205,9 +205,10 @@ - (BOOL) atLeastOneAnimatingOnMainThreadWithQuery:(id) query {
                                                       BOOL *stop) {
             CAAnimation *animation = [view.layer animationForKey:key];
             if (animation.duration > kLPConditionRouteAnimationDurationLimit &&
-               ! [key  isEqual: @"UITextSelectionViewCaretBlinkAnimation"]) { 
+                ! [key  isEqual: @"UITextSelectionViewCaretBlinkAnimation"]) {
```
jamuc, line 208:
```suggestion
                ![key  isEqual: @"UITextSelectionViewCaretBlinkAnimation"]) { 
```

jamuc (approved):

PR #475, opened 2022-02-03
Basic SwiftUI support

JoeSSS:
This PR adds a basic SwiftUI support to calabash server


JoeSSS:
Cannot reproduce the tests failures locally so far, and do not have access to azure to check what the problem is there

JoeSSS:
there is one flaky tests that compares 20 with 20.0000001 in an unrelated part of the app, all the rest tests passing locally.

PR #479, opened 2022-06-28
Apply workaround for predicate assertion issue on iOS 15.5+ versions

agileseph:
From iOS 15.5, it is added security action check that fails NSFunctionExpression with the following error:
    "(Foundation) [com.apple.Foundation:general] NSPredicate: NSFunctionExpression with selector 'accessibilityLabel' is forbidden."
The issue is related to any accessibility property. However, it is used identifier and label properties in tests. In this pull request, I read accessibility id or label values first and then evaluate the predicate with these values.


agileseph:
Hi, @JoeSSS , could u please have a look at the PR?

JoeSSS (approved):
Lgtm

IlyaBausovAkvelon:
@agileseph is it still happening in the iOS 16?

